PR #14, opened 2023-09-17
added wait for action node

gvrocksnow:
Added a logic node for waiting until an InputEventAction is pressed.


Naros:
Hi @gvrocksnow, congrats on your very first contribution 🎉 .  This new node looks like a great addition, I'll add a few commits shortly for documentation and this should be good.

PR #60, opened 2024-01-06
GH-59 Support border/corner radius & gradient title bars

### LINKED ISSUE #59: Visual Improvements - Border/Corner Radius & Gradient Titlebars
### Description

As discussed on Discord, it would be neat to add additional cosmetic features allowing users to specify a custom border/corner radius that is applied to the `GraphNode` instances and optionally specify gradient title bars.

**Border/Corner Radius**

We probably should limit this to a range, perhaps 0-24 or maybe even 0-16. Going above those upper bounds will likely distort the application's visual fidelity just due to how these work.  Additionally, border/corner radius values cannot be applied to gradient title boxes. This is a Godot limitation with how StyleBox objects work.

**Gradient Title Bars**

Gradient title bars work by applying a `StyleBoxTexture2D` rather than a `StyleBoxFlat` theme to the `GraphNode`. These do not have the same features as the flat style box, particularly, they exclude the ability to set border/corner radius and border colors. This means they don't work well with the border/corner radius concept above, nor will they render well in selection highlighting, which is currently done by shading the node's border with yellow.

For now, when Gradient title bars are enabled, and a node is selected, the Gradient effect is disabled on the node, and it will render as though it's using the flat style box theme so that the selection looks reasonable without shifting the title bar's color to compensate, which may be worse for accessibility reasons.

### Implementation ideas

We need to add two new settings to `OrchestratorSettings`, specifically:

* `ui/nodes/titlebar/use_gradient/colors`
* `ui/nodes/border_radius`

For gradient colors, this will be of the type `Variant::BOOL`, which will either toggle the behavior on or off. The style box will be created using a `StyleBoxTexture2D` with an inverted, linear gradient texture based on the node's color defined elsewhere in the settings panel.  This option should default to `false`.

For border-radius, this will be a `Variant::INT` based on a range of `0-24` with a step range of `1`.   This option will default to `6`, slightly improving the current standard border/corner radius.

Naros:


src/editor/graph/graph_node.cpp
```diff
@@ -379,26 +413,98 @@ Ref<StyleBoxFlat> OrchestratorGraphNode::_make_colored_style(const String& p_exi
...
+
+    Ref<GradientTexture2D> texture(memnew(GradientTexture2D));
+    texture->set_gradient(gradient);
+    texture->set_width(64);
+    texture->set_height(64);
+    texture->set_fill_to(Vector2(1.1, 0));
+
+    Ref<StyleBoxTexture> titlebar_tex(memnew(StyleBoxTexture));
+    titlebar_tex->set_texture(texture);
+    titlebar_tex->set_modulate(p_color);
```
Naros, line 464: Rather than apply the color on the `modulate` function, we should apply that to override the white component in the above `PackedColorsArray`.  Additionally, the user should be able to supply a secondary Gradient color in the settings that would effectively replace the black. By default, the second color would be `#000000`.

PR #218, opened 2024-03-31
GH-217 Add signal connection indicator for user-defined functions

### LINKED ISSUE #217: Highlight signals connected to user-defined functions
### Description

For event-graph-based functions, we illustrate when signals are connected, as shown here:
image

However, there is a signal connected to `new_game`, but no icon indicates that it is connected. 

### Implementation ideas

_No response_

Naros:
Fixes #217

### DIFF
--- a/src/common/scene_utils.cpp
+++ b/src/common/scene_utils.cpp
@@ -133,7 +133,7 @@ namespace SceneUtils
     Vector<Node*> find_all_nodes_for_script(Node* p_base, Node* p_current, const Ref<Script>& p_script)
     {
         Vector<Node*> nodes;
-        if (p_current->get_owner() != p_base && p_base != p_current)
+        if (!p_current || (p_current->get_owner() != p_base && p_base != p_current))
             return nodes;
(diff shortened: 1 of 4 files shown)


src/common/scene_utils.cpp
```diff
@@ -147,4 +149,48 @@ namespace SceneUtils
         return nodes;
     }
 
+    Vector<Node*> find_all_nodes_for_script_in_edited_scene(const Ref<Script>& p_script)
+    {
+        SceneTree* scene_tree = Object::cast_to<SceneTree>(Engine::get_singleton()->get_main_loop());
+        if (scene_tree)
+        {
+            Node* scene_root = scene_tree->get_edited_scene_root();
+            return find_all_nodes_for_script(scene_root, scene_root, p_script);
```
Naros, line 158: Need to protect against a `null` scene root dispatch.

Naros, line 158: Need to protect against a `null` scene root.

PR #219, opened 2024-03-31
GH-207 Change orchestration view based on active root scene node

### LINKED ISSUE #207: Same behavior when selecting a scene and selecting a script
### Description

If you have GDScript and you have script attached to the root node and you select that scene then the script will be selected if you are in the script editor. It would be benificial if the same behavior is in Orchestrator.

### Implementation ideas

_No response_

Naros:
Fixes #207


Naros:

PR #241, opened 2024-04-07
GH-240 Standardize Method return value check

### LINKED ISSUE #240: get_setting returning null in all scenarios
### Describe the bug

The get_setting node in the ProjectSettings singleton constantly returns `null`.

### Expected behavior

get_setting should normally return the value of the setting, accessed by inputting the path to it.

image

### Actual behavior

Currently, it returns `null` in all cases.

### How to Reproduce?

_No response_

### Godot full version

4.2

### Orchestrator version

v2.0-stable

### Additional information

_No response_

Naros:
Fixes #240


Naros:

PR #244, opened 2024-04-08
GH-239 Provide `ExtensionDB::is_class_enum_bitfield` look-up

### LINKED ISSUE #239: Cannot detect if a class-specific enums is a bitfield
### Description

In the `ClassDB` and the plug-in's `ExtensionDB`, there is no logical way to detect a class-specific enum and to check if that enum is a bitfield. This works against being able to render these as an option in the new variable support subsystem.

Examples include:
- `Control.SizeFlags`
- `FileAccess.UnixPermissionFlags`
- `Node.ProcessThreadMessages`
- `ResourceSaver.SaverFlags`
- and more

### Implementation ideas

These enumerations are available in the `ClassDB`; the issue is there isn't a toggle that indicates if the enum information is for a bitfield or if it should be treated as a regular enum. We get this information if we need to render this as a property for a function call; however this information isn't available when we need to look at enums outside that context.

Therefore, we only need to create a small data structure in the `ExtensionDB` that holds a list of class-specific enum bitfield names, and we can use this for rendering purposes, i.e. `HashMap<StringName, List<StringName>> _class_bitfields` or similar, sourcing this data from the `extension_api.json` file from godot-cpp.

Naros:
Fixes #239

### DIFF
--- a/cmake/extension_db.cpp.in
+++ b/cmake/extension_db.cpp.in
@@ -158,5 +158,12 @@ namespace godot
     {
         return ExtensionDB::_singleton->_functions[p_name];
     }
+
+    bool ExtensionDB::is_class_enum_bitfield(const StringName& p_class_name, const String& p_enum_name)
+    {
+        if (ExtensionDB::_singleton->_classes.has(p_class_name))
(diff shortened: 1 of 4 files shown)


Naros:

PR #246, opened 2024-04-08
GH-245 Force `ExtensionDB` generator to use LF line endings

### LINKED ISSUE #245: ExtensionDB generator uses CRLF line endings on Windows
### Describe the bug

The `api/extension_db.h` and `api/extension_db.cpp` files are generated by the ExtensionDB cmake script, and this cmake extension is using the native operating system line endings, which is `CRLF` on Windows.  Since git translates this file's line endings to LF, when the file gets regenerated the IDE reports that there are changes in the two files, but these changes are only line endings.

### Expected behavior

Regeneration on Windows or Linux should always result in no modifications to `ExtensionDB` files.

### Actual behavior

Modifications detected which are only line endings.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

This can be solved by adding `NEWLINE_STYLE LF` to the end of the CMake `CONFIGURE_FILE` macros in `godot-extension-db-generator.cmake` file.

Naros:
Fixes #245


Naros:

PR #251, opened 2024-04-15
GH-238 Rework/Improve Bitfield pin value rendering

### LINKED ISSUE #238: Bitfield rendering logic flawed with certain bitfields
### Describe the bug

For example, adding a `KeyModifierMask` bitfield as an input type does not correctly handle bitfield mask toggles. This is because we treat bitfield options as single bit toggles, like how `KEY_MASK_CTRL`, `KEY_MASK_ALT`, or `KEY_MASK_SHIFT` work.  

However, there are other values in the bitfield like `KEY_CODE_MASK` and `KEY_MODIFIER_MASK` that are used in logic to get the underlying key code or the modifier respectively, and basically turn off bits to the left or right of the value to get the group of values without the other's bits. Since there is overlap we end up when checking `KEY_MASK_SHIFT`, the `KEY_CODE_MASK` is also enabled, as follows:

image

From a bitwise point of view, this makes sense, but it would confusing for someone who isn't familiar with bitmasks.  

### Expected behavior

Only the user-toggled bitmask options should be toggled.

### Actual behavior

If there are bit-overlaps in the masks, multiple bits can be toggled.

### How to Reproduce?

_No response_

### Godot full version

4.2.1-stable

### Orchestrator version

2.1.dev1

### Additional information

For Orchestrator bitfields, we should look at inspecting the value for each bitfield item, and if it sets more than a single bit in the bitmask, then those would be hidden by default. The UI could allow the user to see all bitfield values if needed; however, this would at least avoid the initial confusion for users.

We should likely review each bitfield in Godot and see whether or not there are any potential gaps with this approach.

Naros:
Fixes #238

### DIFF
--- a/src/editor/graph/pins/graph_node_pin_bitfield.cpp
+++ b/src/editor/graph/pins/graph_node_pin_bitfield.cpp
@@ -18,11 +18,14 @@
 
 #include "api/extension_db.h"
 #include "common/scene_utils.h"
+#include "common/string_utils.h"
 
 #include <godot_cpp/classes/button.hpp>
 #include <godot_cpp/classes/check_box.hpp>
 #include <godot_cpp/classes/grid_container.hpp>
+#include <godot_cpp/classes/h_separator.hpp>
 #include <godot_cpp/classes/popup_panel.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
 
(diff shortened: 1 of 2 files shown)


Naros:

PR #256, opened 2024-04-17
GH-254 Fix emitting non-Orchestrator based signals

### LINKED ISSUE #254: Error - "No signal with name" when trying to emit a signal
### Describe the bug

One concept that games may often use to simplify communication is the idea of an **EventBus**, where an Autoload exposes any number of signals that various objects in your game connect and use to be notified when another object wishes to inform observers of this change.

When a global signal is defined in an Orchestration that acts as an Autoload, dragging from the Autoload shows an action:
```
* Class
  * Signals
    * Node
      * Emit "Signal"
```

The issue is when selecting this action, an error is output in the **Ouput** panel:
* No signal with name SignalName

It's worth noting that even trying to emit signals that are part of the Orchestration's base type, i.e. extends `Button`, selecting `Emit Pressed` also fails with the same error message.

### Expected behavior

No errors

### Actual behavior

Error reported

### How to Reproduce?

1. Create an Autoload Orchestrator
2. Define a signal in the Autoload
3. In a separate script attached to a node, get the autoload
4. Dragging from the autoload option, select the `Emit XXXX` signal action

### Godot full version

4.2.1-stable

### Orchestrator version

2.0.stable

### Additional information

_No response_

Naros:
Fixes #254


Naros:

PR #264, opened 2024-04-21
GH-263 Support text-based resource formats

### LINKED ISSUE #263: Support storing Orchestration resources as text
### Description

Orchestrator defaults to using binary-based formats for `Orchestration` resources; however, this has a drawback in that these tend to be hard on VCS systems and for the developers to know what changed across commits. For this reason alone, it would be helpful to store `Orchestration` resources as text resources rather than binary.

### Implementation ideas

For Orchestrator 2.0 and likely 2.1, we want to allow users to continue using the binary formats by default, particularly if they have invested in that format's use in their games.  To allow for this in a non-disruptive way, a new plug-in setting should be introduced that controls whether or not to use text or binary format by default.  This can be changed in the `ProjectSettings`.

When set to use binary by default, users should notice no difference.  The **Create Script Dialog** will default to using the `.os` extension, and the resources will be stored in binary format.  If a user wants to export or save the script as a text-based `.torch` file, they can use the `File > Save As` menu option from the Orchestrator plug-in panel's main window.

The text-based format will use the same serialization technique that Godot utilizes for `.tscn` and `.tres` resources. While this format is somewhat verbose, it has a significantly smaller footprint than `.xml` and relatively smaller than `.json` or `.yaml`.  The major downside to this is that there is a significant amount of parser logic copied from the Godot engine to support these formats, although maybe some discussion can be explored with Godot to make these available to GDExtension users.

Not only will this change require some refactor work and the new Loader/Saver implementations, a new editor export plug-in will be needed to convert resources during the export stage at a minimum from the text-based storage format to the binary format.  

While this may be out of scope initially, this sets the stage for a new "export-only" format that contains a byte-code compiled version of the script, that has the best performance and optimizations for exported games.

Naros:
Fixes #263 

- [x] Finish editor export plugin changes
- [x] Test in-editor and exported builds
- [x] Update documentation - https://github.com/Vahera/godot-orchestrator-docs/pull/29


Naros:

PR #271, opened 2024-04-21
GH-253 Include registered autoloads in all actions menu

### LINKED ISSUE #253: Include Autoloads in All Actions
### Description

Since an autoload requires being registered in the `Project Settings`, we could potentially simplifying the creation of an autoload accessor by making them directly accessible in the **All Actions** dialog window much like what is done for custom signals, variables, and functions.

### Implementation ideas

So rather than adding a `Get Autoload` node, then using the **Inspector** to change the autoload reference, there would be a section in the **All Actions** window that has:

```
* Autoloads
  * Get GameManager Autoload
```

Selecting the `Get GameManager Autoload` would drop a `Get Autoload` node that pre-selects the `GameManager`.

Naros:
Fixes #253

### DIFF
--- a/src/editor/graph/actions/action_menu.cpp
+++ b/src/editor/graph/actions/action_menu.cpp
@@ -27,6 +27,7 @@
 
 #include <godot_cpp/classes/check_box.hpp>
 #include <godot_cpp/classes/v_box_container.hpp>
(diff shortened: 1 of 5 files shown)


Naros:

PR #274, opened 2024-04-21
GH-248 Construct `Callable` with correct target reference

### LINKED ISSUE #248: Call connect dosn't work
### Describe the bug

Calling connect dosn't listen to that signal and call the method on the connected callable.
If I use make callable I get error in the output
`_call_internal: Script call error #2 with Node 9: Failed to parse expression: Callable(Score:, "_on_update_score")
  <C++ Source>   /home/runner/work/godot-orchestrator/godot-orchestrator/src/script/instances/script_instance.cpp:799 @ _call_internal()`

If I make a gd file with the callable instead the error is removed but the connection still don't call the callable method.
If I make a gd file that look like this
`func connecting(fromNode: Object, fromSignal: String, toNode: Object, toFunction: String):
    fromNode.connect(fromSignal, Callable(toNode, toFunction))`
image
I can connect to this node and everything is working. 

So it looks like there are some error with call connect and maybe also Make Callable

### Expected behavior

Dynamicly connect two scene with signals

### Actual behavior

_No response_

### How to Reproduce?

Connect it like this
image

### Godot full version

4.2.1.stable

### Orchestrator version

v2.0.stable.official [98f58dca8]

### Additional information

_No response_

Naros:
Fixes #248

### DIFF
--- a/src/script/nodes/data/compose.cpp
+++ b/src/script/nodes/data/compose.cpp
@@ -152,6 +152,16 @@ public:
                     case Variant::STRING_NAME:
                         p_context.set_output(0, VariantUtils::cast_to<StringName>(p_context.get_input(0)));
                         break;
+                    case Variant::CALLABLE:
+                    {
+                        Object* callable_object = p_context.get_input(0);
+                        if (!callable_object)
+                            callable_object = _instance->get_owner();
+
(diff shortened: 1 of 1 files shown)


Naros:

PR #278, opened 2024-04-22
GH-206 Correctly generates action items for instantiated scenes

### LINKED ISSUE #206: Instantiated scene don't reflect function scripts
### Describe the bug

If an instantiated scene has script attached to it that has function then it's not reflected in the add node menu.
I have a function called initialize in my script file on my mob. 
image
I instantiate my mob file 
image
When I try to use that one the function is no where to be found
image
If I'm in my mob file it's get found
image

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2.1.stable

### Orchestrator version

v2.0.rc2

### Additional information

_No response_

Naros:
Fixes #206 

This ended up being trickier than expected because for `Instantiate Scene` nodes, we create an instance of the `Scene` and return it to the caller; however, this needed some memory management to work correctly.  We have yet to encounter this with any other node because other objects owned what was returned.

However, this also highlighted one minor issue in the **All Actions** window: when you drag from a pin and search for actions, the list was being regenerated on every key press, which created sluggish responses.  This was fixed with a slight variation to the cache logic so that the list is generated once per "opening" and then reused for each keypress, mimicking the behavior when simply right-clicking the graph.


Naros:

PR #294, opened 2024-04-24
GH-283 Support method chaining

### LINKED ISSUE #283: Call function nodes should have optional target output pin
### Description

There are several types of **Call Function** nodes, and one in particular accepts a `Target` which is the object the method will call on.  In cases where you need to call multiple methods on the target object, this leads to some weird connections.

Instead, it might prove helpful to have these **Call Function** nodes to have an output pin that passes the input `Target` value to the output, following a more conceptual method chaining scenario in text-based languages.

The **Inspector** can have a toggle that enables this output pin per node.

### Implementation ideas

_No response_

Naros:
Fixes #283

### DIFF
--- a/src/editor/graph/graph_edit.cpp
+++ b/src/editor/graph/graph_edit.cpp
@@ -1241,7 +1241,7 @@ void OrchestratorGraphEdit::_on_paste_nodes_request()
     HashMap<int, int> bindings;
     for (const KeyValue<int, Ref<OScriptNode>>& E : _clipboard->nodes)
     {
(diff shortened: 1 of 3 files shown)


Naros:
Few bugs identified:

- [ ] Unchecking the `Chain` option does not disconnect the connection, leading to GraphEdit/GraphNode errors.
- [ ] Duplicating a chained node using `Ctrl+D` works, Copy-n-Paste does not copy the `Chain` state.

Naros:

PR #295, opened 2024-04-24
GH-282 Explicitly filter `_get` and `_set` methods

### LINKED ISSUE #282: Consider hiding `_set` and `_get` overridable methods
### Description

Godot objects provide a `set` and `get` method, used for assigning and getting property values.  There are also virtual methods that do precisely the same, `_set` and `_get`; these are generally called directly by the engine and should not be called directly.  This has caused some confusion for users as its not obvious which of the two are being picked.

### Implementation ideas

Either hide these entirely or consider hiding them unless context sensitivity is disabled.

Naros:
Fixes #282


Naros:

PR #296, opened 2024-04-24
GH-281 Improve rendering of math nodes in all actions

### LINKED ISSUE #281: Use more descriptive mathematical action item names
### Description

Presently, mathematical and logical operation nodes use a confusing description:

image

In this screenshot, this implies the first argument is a `Float` while the second will be a `Color`, `Float`, `Int`, `Quaternion`, etc.  

### Implementation ideas

It may be more beneficial to change the descriptions to read something more natural like:

* Multiply Float and Color
* Multiply Float and Float
* Multiply Float and Integer
* Multiply Float and Quaternion

At least this would give the user a better grasp of the two operands in the operation.

### LINKED ISSUE #1150: Allow cumulative math operations to dynamically add pins
### Description

Specific mathematical operations, such as Addition, Multiplication, bitwise AND, and bitwise OR, allow the addition of operands in any order due to the cumulative nature of the operation, and yield the same result.

This is an effective way to reduce node bloat for specific operations by inputting multiple values into a single node and having it handle the operation's associativity.

### Implementation ideas

Currently, if an operation needs to compute `A * B * C`, it requires two multiply nodes, with the result of the first two operands fed into the second as input, and the final operand. This produces an associative operation `(A * B) * C`. But given that associativity and cumulativity semantics on multiplication do not influence the outcome, this can be reduced down to an `Input1 * Input2 * Input3 * .... InputN` operation.

1. Have the operator node extend the editable pin interface. 
2. Have the interface determine whether the `[+]` button is available based on the node's operator type. For example, addition and multiplication work, whereas subtraction and division do not.

Naros:
Fixes #281 
Fixes #1150 

Images shared on Discord for this feature:
image
image
image
image
image
image

### DIFF
--- a/doc_classes/OrchestratorSettings.xml
+++ b/doc_classes/OrchestratorSettings.xml
@@ -13,6 +13,9 @@
         <member name="editor/settings/updates/notify_about_preview_releases" type="bool" setter="" getter="" default="false">
             When enabled, the updater also reports preview (pre-release) builds in addition to stable releases.
         </member>
+        <member name="editor/behavior/general/enable_type_promotion" type="bool" setter="" getter="" default="true">
+            When set to [code]true[/code], new dynamic and promotable operator nodes are available in the action dialog menu. When set to [code]false[/code], the old non-dynamic operator nodes are available.
+        </member>
         <member name="editor/behavior/files/autosave_interval_secs" type="float" setter="" getter="" default="0.0">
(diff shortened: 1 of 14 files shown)


Naros:
So based on discussions on Discord, it seems the consensus would be to try and design these in a more generalized way, so that users can place a simple Multiply node and modify the types as needed.

PR #309, opened 2024-05-01
GH-306 Fix functions to return Any types

### LINKED ISSUE #306: Cannot return variant type
### Describe the bug

When selecting Any for return type for a function, it implicitly sets has_return_value to false

### Expected behavior

I expected it to simply show up Any parameter instead. Same for when you set has_return_value to true. It implicitly sets it to something that isnt Any.

### Actual behavior

When you set return type to any, has_return_value sets to false
When return type is set to any and you set has_return_type to true, it sets return type to the first thing after any, which is bool.

### How to Reproduce?

_No response_

### Godot full version

v4.3.dev5.official [c9c17d6ca]

### Orchestrator version

v2.0

### Additional information

_No response_

Naros:
Fixes #306

### DIFF
--- a/src/common/method_utils.cpp
+++ b/src/common/method_utils.cpp
@@ -32,4 +32,26 @@ namespace MethodUtils
         // No return value
         return false;
     }
+
+    void set_no_return_value(MethodInfo& p_method)
+    {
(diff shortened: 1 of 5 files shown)


Naros:

PR #324, opened 2024-05-03
GH-323 Fix enum persistence as a default value

### LINKED ISSUE #323: Class enum default values not saved
### Describe the bug

When a node uses a class-based enum as an input argument, such as `FileAccess:open`, the enum state is not managed properly and is reset when the orchestration is closed or the editor is restarted.

### Expected behavior

Default value state persisted.

### Actual behavior

Default value state is lost for enum pins based on class-specific enums.

### How to Reproduce?

_No response_

### Godot full version

4.2.1-stable

### Orchestrator version

2.0.stable

### Additional information

_No response_

Naros:
Fixes #323


Naros:

PR #329, opened 2024-05-04
GH-221 View doc jump to class/constant/method Godot doc

### LINKED ISSUE #221: Allow jumping directly to a Godot function call's documentation
### Description

For example, given the following Call function, which calls `start()` on a `Timer`:
image

The context menu has a `View Documentation` option, it would be more useful if either Orchestrator was able to render information about the function, its signature, and description.  If that isn't possible due to GDExtension limitations, would it be at least possible to refer the user to the actual function documentation as there should be sufficient context to redirect there, i.e.:

image

### Implementation ideas

_No response_

Naros:
Fixes #221

### DIFF
--- a/src/common/method_utils.cpp
+++ b/src/common/method_utils.cpp
@@ -16,6 +16,8 @@
 //
 #include "common/method_utils.h"
 
+#include <godot_cpp/core/class_db.hpp>
+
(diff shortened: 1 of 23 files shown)


Naros:
Applied to main (=2.1).

PR #336, opened 2024-05-07
GH-156 Collapse nodes to function and expand node feature

### LINKED ISSUE #156: Select node and easy move them to a function
### Description

The graph will get big kinda fast. You can move nodes to function and call them from the main script.

### Implementation ideas

Be able to select nodes and right click or in some way create a new function with the nodes we have selected. Create the new function and have the chain uninterupted in the graph. If there are function argument and return values these should be created as well.

Naros:
Fixes #156 

Given a graph like this one:

image

There is a new `Organization` section on the node context menu:

image

With nodes selected, pick "Collapse to Function" to move the selected nodes to a new user-defined function:

image
image

On the context-menu, a user-defined script function can be expanded, which will replace the call script function node with the nodes from the child function, as follows:

image

When expanding the following are not performed and left as exercises for the user:

1. Previous connections to the call function node are not preserved.
2. The function is not removed automatically; it's left in case it is used in other places.

### DIFF
--- a/src/common/callable_lambda.h
+++ b/src/common/callable_lambda.h
@@ -1,0 +1,113 @@
+// This file is part of the Godot Orchestrator project.
+//
+// Copyright (c) 2023-present Vahera Studios LLC and its contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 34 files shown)


Naros:

PR #351, opened 2024-05-12
GH-350 Add for-each/for-loop abort output pin

### LINKED ISSUE #350: For Each/Loop - Abort output flow pin
### Description

To identify if a for-loop for for-each-loop broke, users must check the state that triggered the loop's body.  Since the loop node knows that the broken input pin was fired, this can be passed as an output pin so that users can easily deviate logic on completed vs aborted.

### Implementation ideas

_No response_

Naros:
Fixes #350


Naros:

PR #361, opened 2024-05-15
GH-360 Select action names from drop-down list

### LINKED ISSUE #360: Render `Input` and `InputEvent` action parameters as drop-down
### Description

The `Input` and `InputEvent` object has several methods that take an action name. When the call function's target pin is linked to either of these source types, we should render the action list as a drop-down of all Godot and user-defined action types.

### Implementation ideas

_No response_

Naros:
Fixes #360


Naros:

PR #375, opened 2024-05-27
GH-365 Add disconnect control flow on drag option

### LINKED ISSUE #365: Control flow pins should break when initially dragged
### Description

When an execution pin is linked with another pin on another node, if the user drags away from the execution pin, it will begin to draw a secondary connection. Eventually, if that connection is linked to another node, the original execution control flow will be removed as control flow execution is only allowed to advance sequentially.

image

Some users would like to prevent this as it may seem confusing and would instead prefer the original pin to be broken upon the initial drag operation.  This could potentially be accomplished a few different ways:

1. Investigate the current base options of `GraphEdit`; currently, we use a break when dragging on the right.
2. Use a modifier on the initial drag to immediately break the connection, i.e., `CTRL`.

We may want to consider making this togglable via some configuration option(s) to customize this behavior.

### Implementation ideas

_No response_

Naros:
Fixes #365

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -143,6 +143,7 @@ void OrchestratorSettings::_register_settings()
 
     _settings.emplace_back(BOOL_SETTING("ui/actions_menu/center_on_mouse", true));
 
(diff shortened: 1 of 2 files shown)


Naros:

Issue #367, opened 2024-05-19
Make key bindings configurable


Naros:
### Description

The editor has an `EditorSettings` section where short-cuts can be configured.  Can Orchestrator submit key bindings and allow the user to configure those here?

Naros:
The `Editor > Editor Settings > Shortcuts` are not accessible to GDExtension unfortunately.  In addition, there are bugs where adding custom actions to `InputMap` do not refresh the Godot Editor, causing the entries to be invisible until the editor is restarted.  // cc @eobet

Naros:
Godot 4.6 has introduced the ability to add, get, and remove shortcuts from the EditorSettings UI based on what I can see in Godot's source, so it should be feasible to introduce this in Orchestrator 2.4 (based on Godot 4.6+).

bit-garden:
I'm not sure what UE keybinds were there. I personally would like quick node creation keybinds. Something like ctrl + number to extend a pin on the right, or ctrl + shift + number to extend a pin on the left. 

Sort out the tab order on some of the fields. (I'm not super sure how this would work)

Ctrl + f for finding a specific node. Ctrl+N or P for jumping to the next possible node in the search?

PR #376, opened 2024-05-27
GH-368 Use Shift+LMB to place knots, Ctrl+LMB to remove

### LINKED ISSUE #368: Consider adjusting the key bind for Place/Remove Knots
### Description

Placing and removing many knots on the graph using the same key binding is likely not ideal.  We should consider using SHIFT+LMB and CTRL+LMB and decide which of the two bindings refers to placement versus removal.

This could go in parallel with #367 so that users have full control over such bindings in the end.

### Implementation ideas

_No response_

Naros:


Naros:
Closed in favor of 
* #367

PR #386, opened 2024-06-07
GH-383 Split out `Orchestration` contract and refactors

### LINKED ISSUE #383: Split the notion of an `Orchestration`
### Description

In the current design, the code treats the concept of an `Orchestration` as an `OScript`.  In other words, when a `.os` file is loaded, the code creates an instance of the `OScript` resource, which by design is an attachable resource on a node in the scene.  But in an effort to extend Orchestrator to support concepts like macro libraries, we need to split this definition.

### Implementation ideas

Conceptually, we should aim for something like this:

### Orchestration
```cpp
class Orchestration {
  // All common code here
};
```

### OScript
```cpp
class OScript : public godot::ScriptExtension, public Orchestration {
  // Implementation of Godot's ScriptExtension
  // Any OScript-specific behavior only
};
```

### OScriptMacroLibrary
```cpp
class OScriptMacroLibrary : public godot::Resource, public Orchestration {
  // Implementation of the Macro Library behavior 
};
```

Most code would then take an `Orchestration` reference rather than an `OScript` or `OScriptMacroLibrary` reference.  This will allow the plug-in to work with the common contract independent of the actual underlying resource loaded and only branch on the case-by-case scenarios.

This has the added value that as we want to extend Orchestrator to support things such as state-graphs, we've already built the architecture to have a single resource file that can contain a combination of state graphs and other orchestration bits or a defined subset of these based on rule-sets.

Naros:
Fixes #383


src/editor/component_panels/signals_panel.cpp
```diff
@@ -106,8 +106,12 @@ Dictionary OrchestratorScriptSignalsComponentPanel::_handle_drag_data(const Vect
     TreeItem* selected = _tree->get_selected();
     if (selected)
     {
-        data["type"] = "signal";
-        data["signals"] = Array::make(selected->get_text(0));
+        Ref<OScriptSignal> signal = _orchestration->find_custom_signal(StringName(selected->get_text(0)));
```
Naros, line 109:
```suggestion
        Ref<OScriptSignal> signal = _orchestration->find_custom_signal(selected->get_text(0));
```
Consider changing the argument type from `StringName` to `String` for simplicity.

src/editor/graph/actions/default_action_registrar.cpp
```diff
@@ -30,7 +30,9 @@ void OrchestratorDefaultGraphActionRegistrar::_register_node(const OrchestratorG
                                                        const Dictionary& p_data)
 {
     OScriptLanguage* language = OScriptLanguage::get_singleton();
-    Ref<OScriptNode> node = language->create_node_from_name(p_class_name, p_context.script, false);
+    Orchestration* orchestration = p_context.graph->get_orchestration();
+
+    Ref<OScriptNode> node = language->create_node_from_name(p_class_name, orchestration, false);
```
Naros, line 35:
```suggestion
    Ref<OScriptNode> node = p_context.graph->create_node_from_name(p_class_name, false);
```
Can we abstract this behind the `Ref<OScriptGraph>` for future expansion needs.

src/editor/graph/graph_edit.cpp
```diff
@@ -391,6 +367,32 @@ void OrchestratorGraphEdit::execute_action(const String& p_action_name)
     Input::get_singleton()->parse_input_event(action);
 }
 
+Ref<OScriptNode> OrchestratorGraphEdit::spawn_node(const StringName& p_type, const OScriptNodeInitContext& p_context, const Vector2& p_position)
+{
+    Ref<OScriptNode> spawned = _script_graph->create_node(p_type, p_context, p_position);
```
Naros, line 372: Can be `const`

src/editor/graph/graph_edit.cpp
```diff
@@ -391,6 +367,32 @@ void OrchestratorGraphEdit::execute_action(const String& p_action_name)
...
+    if (spawned.is_valid())
+    {
+        if (_drag_context.should_autowire())
+        {
+            Ref<OScriptNode> other = _script_graph->get_node(_drag_context.node_name.to_int());
```
Naros, line 377: Can be `const`.

src/editor/graph/graph_edit.cpp
```diff
@@ -1833,8 +1782,8 @@ void OrchestratorGraphEdit::_on_paste_nodes_request()
 
 void OrchestratorGraphEdit::_on_script_changed()
 {
-    _base_type_button->set_button_icon(SceneUtils::get_editor_icon(_script->get_base_type()));
-    _base_type_button->set_text(vformat("Base Type: %s", _script->get_base_type()));
+    _base_type_button->set_button_icon(SceneUtils::get_editor_icon(_script_graph->get_orchestration()->get_base_type()));
+    _base_type_button->set_text(vformat("Base Type: %s", _script_graph->get_orchestration()->get_base_type()));
```
Naros, line 1786: We make a number of calls in this class to `_script_graph->get_orchestration()`, hide this behind a method.

src/editor/script_view.cpp
```diff
@@ -171,8 +154,8 @@ void OrchestratorScriptView::_move_nodes(const Vector<Ref<OScriptNode>>& p_nodes
 {
     for (const Ref<OScriptNode>& E : p_nodes)
     {
-        p_source->remove_node(E->get_id());
-        p_target->add_node(E->get_id());
+        p_source->remove_node(E);
+        p_target->add_node(E);
     }
 }
```
Naros, line 160: Can this be refactored into `OScriptGraph` somehow?

src/orchestration/build_log.cpp
```diff
@@ -0,0 +1,33 @@
...
+void BuildLog::error(const String& p_message)
+{
+    _messages.push_back(vformat("[b][color=#a95853]ERROR[/color][/b]: Node %s ([url={\"goto_node\":\"%d\"}]#%d[/url]) - %s",
+        _current_node->get_class(), _current_node->get_id(), _current_node->get_id(), p_message));
+    _errors++;
+}
+
+void BuildLog::warn(const String& p_message)
+{
+    _messages.push_back(vformat("[b][color=yellow]WARNING[/color][/b]: Node %s (#%d) - %s", _current_node->get_class(), _current_node->get_id(), p_message));
```
Naros, line 30: Add meta logic here for warnings to match errors.

src/orchestration/orchestration.cpp
```diff
@@ -0,0 +1,911 @@
...
+        WARN_PRINT(vformat("Removed orphan node %d from script %s.", E.key, get_self()->get_path()));
+        _nodes.erase(E.key);
+    }
+
+    for (const OScriptConnection& E : _connections)
+    {
+        if (!_nodes.has(E.from_node) || !_nodes.has(E.to_node))
+        {
+            WARN_PRINT("Orphan connection found: " + E.to_string());
+            //_connections.erase(E);
```
Naros, line 177: We print this but don't erase, what's the point?

src/orchestration/orchestration.cpp
```diff
@@ -0,0 +1,911 @@
...
+        p_log.set_current_node({});
+    }
+}
+
+bool Orchestration::has_node(int p_node_id) const
+{
+    return _nodes.has(p_node_id);
+}
+
+void Orchestration::add_node(const Ref<OScriptGraph>& p_graph, const Ref<OScriptNode>& p_node)
```
Naros, line 254: Could `add_node` only exist publically on `OScriptGraph` and then it has friend access to `Orchestration`?

src/orchestration/orchestration.cpp
```diff
@@ -0,0 +1,911 @@
...
+    p_node->_validate_input_default_values();
+
+    // Register the node with the script
+    _nodes[p_node->get_id()] = p_node;
+
+    // Register the node with the graph
+    p_graph->add_node(p_node);
+}
+
+void Orchestration::remove_node(int p_node_id)
```
Naros, line 271: Could `remove_node` only exist publically on `OScriptGraph` and then it has friend access to `Orchestration`?

src/orchestration/orchestration.cpp
```diff
@@ -0,0 +1,911 @@
...
+        results.push_back(E.value);
+    return results;
+}
+
+const RBSet<OScriptConnection>& Orchestration::get_connections() const
+{
+    return _connections;
+}
+
+void Orchestration::connect_nodes(int p_source_id, int p_source_port, int p_target_id, int p_target_port)
```
Naros, line 335: Could `connect_nodes` only exist publically on `OScriptGraph` and then it has friend access to `Orchestration`?

src/orchestration/orchestration.cpp
```diff
@@ -0,0 +1,911 @@
...
+    connection.to_node = p_target_id;
+    connection.to_port = p_target_port;
+
+    ERR_FAIL_COND_MSG(_connections.has(connection), "A connection already exists: " + connection.to_string());
+    _connections.insert(connection);
+
+    _self->emit_signal("connections_changed", "connect_nodes");
+}
+
+void Orchestration::disconnect_nodes(int p_source_id, int p_source_port, int p_target_id, int p_target_port)
```
Naros, line 351: Could `disconnect_nodes` only exist publically on `OScriptGraph` and then it has friend access to `Orchestration`?

src/orchestration/orchestration.cpp
```diff
@@ -0,0 +1,911 @@
...
+{
+    for (const KeyValue<StringName, Ref<OScriptFunction>>& E : _functions)
+    {
+        if (E.value->get_guid() == p_guid)
+            return E.value;
+    }
+    return nullptr;
+}
+
+void Orchestration::rename_function(const StringName& p_old_name, const StringName& p_new_name)
```
Naros, line 635: I wonder if `rename_xxxx` should be done through the XXXX target object and friend delegate to `Orchestration`.  What about other resource types.

src/script/graph.cpp
```diff
@@ -153,76 +257,121 @@ void OScriptGraph::set_flags(BitField<GraphFlags> p_flags)
...
+    // script_view#_add_callback
+    const Ref<OScriptNodeEvent> event_entry = p_node;
+    if (event_entry.is_valid() && _flags.has_flag(GF_EVENT) && !_functions.has(p_node->get_id()))
+        _functions.insert(p_node->get_id());
+
+    emit_signal("node_added", p_node->get_id());
 }
 
-TypedArray<int> OScriptGraph::get_functions() const
+void OScriptGraph::add_node2(const Ref<OScriptNode>& p_node)
```
Naros, line 339: Rework this to avoid this second method.

src/script/graph.cpp
```diff
@@ -153,76 +257,121 @@ void OScriptGraph::set_flags(BitField<GraphFlags> p_flags)
...
+void OScriptGraph::remove_node(const Ref<OScriptNode>& p_node)
 {
-    _functions.clear();
-    for (int i = 0; i < p_functions.size(); i++)
-        _functions.insert(p_functions[i]);
+    _remove_node(p_node->get_id());
+}
 
-    emit_changed();
+void OScriptGraph::remove_node2(const Ref<OScriptNode>& p_node)
```
Naros, line 349: Rework this to avoid this second method.

Naros (commented):
Some inline things to change.

src/orchestration/orchestration.cpp
Naros: There is an order of operations issue with `OScriptNodePin`, that we will need to fix before this can be done.

src/orchestration/orchestration.cpp
Naros: Going to leave this as-is.

src/orchestration/orchestration.cpp
Naros: Leaving this as-is for now.

Naros:

PR #387, opened 2024-06-08
GH-385 Fix instantiate scene to respect input pin

### LINKED ISSUE #385: Instantiate Scene does not honor input scene pin name
### Describe the bug

The node fails to load the scene because it does not map the input pin value to the resolution logic in the runtime instance.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2.2-stable

### Orchestrator version

2.1.dev2

### Additional information

_No response_

Naros:
Fixes #385


Naros:

Issue #261, opened 2024-04-18
Support for creating classes


TBNEO:
### Description

The idea is simple: Addition of creating classes and such within an orchestration, similarly to how, in GDScript, one can add the **class_name [name here]** at the top. 
The reasoning behind why do this is mainly for ease of use in inheritance systems. Currently, there isn't an easy way (that i know of) to make an orchestration extend from another. Example,
**Player.os inherits movement capabilities from: Entity.os** 
However, this can't be done at this moment right now.

### Implementation ideas

In implementation, this could be solved by adding perhaps a bool for whether it's to be class or not, along with a necessary Class Name field, and this could be done in the script inspector for ease. 
Doing this with nodes would be inefficient, I think, but if that's easier, then it works too.

Either way, thanks for reading this, good luck with the project.

Naros:
So here are some findings after looking into what is required to make this work:

- Add serialized properties for `class_name` and `icon`
- `OScript::_get_global_name` should return the `class_name`
- `OScript::_get_class_icon_path` should return the `icon`
- `OScriptLanguage::_get_global_class_name` needs to be implemented (more below)
- `OScriptLanguage::_get_type` should return `OScript` rather than `Orchestrator` (more below)
- For an object to show up in **Create New Node** dialog, cannot extend abstract classes, i.e., `CanvasItem.`
  - _This seriously looks like a bug, as it's legal to extend CanvasItem even in GDScript.  
- `OScript::_get_method_info` should be implemented to return method details for GDScript context look-ups.
- `OScript::_has_static_method` should be implemented to expose static methods to GDScript (out of scope).
- `OScript::_get_documentation` should be implemented for context-help like GDScript (can be follow-up).
- `OScript::_get_script_property_list` should be implemented?

So one issue that I don't particularly care for is how Godot resolves the global class name, specifically in the call to the language function `_get_global_class_name`. This call relies on opening the script, parsing the contents, only to return the following:
```gdscript
Dictionary result;
result["name"] = script->get_global_name();
result["base_type"] = script->get_instance_base_type();
result["icon_path"] = script->get_class_icon_path();
```

[...]

PR #388, opened 2024-06-08
GH-384 Support script documentation

### LINKED ISSUE #384: Generate Orchestration script documentation
### Description

A developer should be able to search for an Orchestration by its filename, much like an unnamed GDScript class and to view its associated documentation like other script languages, i.e.:

image

### Implementation ideas

This requires four changes:

1. `OScriptLanguage::get_type()` should return `OScript::get_class_static()` rather than `Orchestrator`.
2. The language's icon should be registered with `OScript::get_class_static()` in `plugin.cpp`.
3. `OScriptLanguage::_supports_documentation()` should return `true`.
4. `OScript::_get_documentation()` should return the documentation metadata.

Naros:
Fixes #384 

- [ ] There is no way to reload documentation once opened automatically when the script is modified.  See https://github.com/godotengine/godot/pull/92916


Naros:
This was resolved in GH-261, with class support.

PR #389, opened 2024-06-08
GH-382 Avoid recent history duplicates

### LINKED ISSUE #382: Variable type dialog duplicates recent history items
### Describe the bug

When opening the Variable Type dialog and selecting the same int multiple times, the entry is added to the start of the recent history list, which gets shown as follows:

image

Ideally, suppose the item exists already in the recent history. In that case, it should be removed and added as the most recent item only, meaning that the size of the list does not change when selecting an existing item again, but only reorders the recent history based on the selection order the user did.

### Expected behavior

Don't duplicate recently selected items

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.3.beta1

### Orchestrator version

2.0.dev2

### Additional information

_No response_

Naros:
Fixes #382

### DIFF
--- a/src/editor/search/variable_classification_dialog.cpp
+++ b/src/editor/search/variable_classification_dialog.cpp
@@ -24,6 +24,7 @@
 #include "editor/plugins/orchestrator_editor_plugin.h"
 
 #include <godot_cpp/classes/engine.hpp>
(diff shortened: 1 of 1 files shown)


Naros:

PR #395, opened 2024-06-09
GH-394 Split virtual machine from `OScriptInstance`

### LINKED ISSUE #394: Separate the virtual machine runtime from `OScriptInstance`
### Description

In the first implementation of Orchestrator, the runtime virtual machine was embedded as part of the `OScriptInstance`, which is what is responsible for executing the scripts at runtime.  This worked quite well; however, there are several benefits to consider when separating the virtual machine from the instance.

One long-term goal is to replace the `OScriptNodeInstance` implementations with a way that supports exporting the script's code not only to a format that the plug-in can run at runtime, but also to other languages such as GDScript or even C#.  By being able to do this using the same mechanism keeps the technical debt low and eases long-term maintenance.

Another benefit is sandboxing.  In some engines they perform an expand/flatten step on the graph-side.  The parts that would need to be sandboxed are expanded.  For example, in UE, this is precisely what happens when macro references.  In the long-term, this is precisely what we'd like to do, but this requires some additional reconstructing of the data model that would be more disruptive than necessary right now.  

The other alternative is to scope parts of the graph, creating scope boundaries around the parts that should be sandboxed.  Using the macro example, in this case the actual script would be scoped to a VM and each macro would have its own VM instance.  This avoids the need to duplicate things on the graph side and leaves the expansion step solely at the runtime side for now, achieving the same outcome as doing graph-side expansion.

### Implementation ideas

_No response_

Naros:
Fixes #394


Naros:

PR #396, opened 2024-06-09
GH-366 Permit deleting event nodes like any other node

### LINKED ISSUE #366: Allow Event nodes to be removed directly from the graph.
### Description

Currently, the `OScriptNodeEvent` node types extend the `OScriptFunctionalEntryNode` primarily to handle a lot of the underlying function pin logic universally; however, the `OScriptFunctionEntryNode` does not allow you to directly delete it from the graph because it is mandatory for a function.

We primarily restrict the deletion of the `OScriptNodeEvent` because we need to coordinate this with the `OScriptGraph`, but perhaps the code can be reworked so that the event nodes can be removed from the graph just like any other node?

### Implementation ideas

_No response_

Naros:
Fixes #366


Naros:

PR #397, opened 2024-06-09
GH-364 Improved autowire with selection dialog

### LINKED ISSUE #364: Re-enable target auto-wire
### Description

Previously, target auto-wiring was disabled due to some functions accepting both a target and another object, and there wasn't an adequate context for the auto-wire system to determine which input should be wired.  In response to this, we disabled auto-wiring on nodes that have `Target` pins.  

### Implementation ideas

When dragging from an output pin, we should _always_ attempt to auto-wire, including the `Target` input pin.  

One way we could combat the problem such as `add_child` where both a parent node (`Target`) and the node to be added (`Child`) are part of the function signature, the user could be presented with a dialog box with all possible input pins that accept that type, and the user selects which pin to auto-wire. This dialog should only be shown if multiple input pins match the output pin's compatibility matrix.

Naros:
Fixes #364

### DIFF
--- a/src/editor/graph/autowire_selections.cpp
+++ b/src/editor/graph/autowire_selections.cpp
@@ -1,0 +1,109 @@
+// This file is part of the Godot Orchestrator project.
+//
+// Copyright (c) 2023-present Vahera Studios LLC and its contributors.
+//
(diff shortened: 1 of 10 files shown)


Naros:

PR #400, opened 2024-06-11
GH-399 Spawn function return node when function has return value

### LINKED ISSUE #399: Spawn return node when setting function has a return value
### Description

The Inspector is where a user will set whether a function returns a value.  When the `Has Return Value [ ]` option is checked, if the graph currently does not have a return value node, the return value node should be automatically spawned.  Additionally, if the entry node's execution pin is not connected to any node, the entry execution pin should be connected to the return's execution pin.

### Implementation ideas

_No response_

Naros:
Fixes #399


Naros:

PR #404, opened 2024-06-15
GH-403 Remove redundant/superfluous text from node/pin names

### LINKED ISSUE #403: Remove redundant / superfluous text from node titles & pins
### Description

The create node dialog & node titles make heavy use of the term `Call`, so it loses its significance.  Additionally, several pins could use some improvements to reduce the use of `Return Value` where applicable.

https://discord.com/channels/1049457206648655912/1240066032777822381/1240066032777822381

### Implementation ideas

_No response_

Naros:
Fixes #403


Naros:

PR #411, opened 2024-06-16
GH-410 Add friendly names for Graph/Function component panels

### LINKED ISSUE #410: Allow friendly names in the component panel
### Description

The component panel renders graphs and functions using more technical terms.  For example, a function called `get_movement` is shown in the component panel using the underscore; however, in the node graph, that function is rendered with the title `Get Movement`.  This can introduce inconsistency or even confusion for some.

### Implementation ideas

Two new UI settings will be introduced in `Project Settings > Orchestrator > UI > Component Panel`:

image

When enabled, the graph & function panels will render the elements using Godot's `capitalize` function, which aims to present more user-friendly names by replacing underscores with spaces and capitalizing each word.

image

When these UI settings are not enabled (the default), the panel uses its current render behavior:

image

In addition, the component panel uses the `PlayStart` icon for event nodes and it uses the `MemberMethod` icon for functions.  The current nodes in the graph canvas do not use the same icons, these should be aligned to the component panel, like this:

image

Naros:
Fixes #410 

- [x] The consensus is to enable friendly names by default.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -143,6 +143,9 @@ void OrchestratorSettings::_register_settings()
 
     _settings.emplace_back(BOOL_SETTING("ui/actions_menu/center_on_mouse", true));
 
+    _settings.emplace_back(BOOL_SETTING("ui/components_panel/show_graph_friendly_names", false));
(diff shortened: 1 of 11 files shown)


Naros:

PR #415, opened 2024-06-17
GH-414 Fix method exclusion logic to allow virtual methods

### LINKED ISSUE #414: Function overrides no longer work
### Describe the bug

A regression was introduced when excluding functions that start with an underscore.  This should not exclude virtual methods.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #414


Naros:

PR #417, opened 2024-06-17
GH-416 Support instantiation of script-based classes

### LINKED ISSUE #416: Orchestation cannot create new objects defined in GDScript
### Describe the bug

Assuming a GDScript object has been defined as follows:
```python
extends Node
class_name TestObject
```

Then in the Orchestration, use a `New Object` node to create an instance of `TestObject`.  
The runtime will output a debugger error:

```
E 0:00:00:0817   instantiate: Cannot get class 'TestObject'.
  <C++ Error>    Parameter "ti" is null.
  <C++ Source>   core/object/class_db.cpp:358 @ instantiate()
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2.2-stable

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #416

### DIFF
--- a/src/script/nodes/memory/memory.cpp
+++ b/src/script/nodes/memory/memory.cpp
@@ -20,20 +20,46 @@
 
 #include <godot_cpp/classes/engine.hpp>
 #include <godot_cpp/classes/node.hpp>
+#include <godot_cpp/classes/project_settings.hpp>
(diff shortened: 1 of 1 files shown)


Naros:

PR #420, opened 2024-06-18
GH-255 Apply property usage fixup to functions/signals

### LINKED ISSUE #255: Orchestrator Signal and Functions are initialized with incorrect usage flags
### Describe the bug

The `PropertyInfo` struct initializes the `usage` property as `7` rather than as `PROPERTY_USAGE_DEFAULT`, which is `6`.  This seems to be a disconnect between GDExtension's `godot-cpp` source and the upstream Godot Engine changes made in 2022.

### Expected behavior

Signal and Function arguments and return values should be initialized with `6` (PROPERTY_USAGE_DEFAULT).

### Actual behavior

Signal and Function arguments and return values should be initialized with `7` (PROPERTY_USAGE_DEFAULT + 1).

### How to Reproduce?

Create any Signal or Function and look at its return value or method arguments.

### Godot full version

4.2.1-stable

### Orchestrator version

2.0.stable

### Additional information

Orchestrator can prevent this bug by simply ensuring that the `PropertyInfo` _usage_ property is initialized using the engine's `PROPERTY_USAGE_DEFAULT` value rather than relying on the struct's initial value.

We'll likely need to define a clean-up hook for existing Orchestrations to sanitize the `OScriptFunction` and `OScriptSignal` definitions.  This should not affect standard method data from the serialized engine.

Naros:
Fixes #255

### DIFF
--- a/src/script/function.cpp
+++ b/src/script/function.cpp
@@ -66,6 +66,18 @@ bool OScriptFunction::_set(const StringName &p_name, const Variant &p_value)
     {
         _method = DictionaryUtils::to_method(p_value);
         _returns_value = MethodUtils::has_return_value(_method);
+
+        // Cleanup the argument usage flags that were constructed incorrectly due to godot-cpp bug
+        for (PropertyInfo& argument : _method.arguments)
+        {
+            if (argument.usage == 7)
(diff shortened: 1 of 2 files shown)


Naros:

PR #424, opened 2024-06-19
Add/Align usage of static `_bind_methods` in node classes

Naros:
* Adds missing `_bind_methods()` method for several nodes (compile-time check added in Godot 4.3)
* Fixes several macro usages of `ORCHESTRATOR_NODE_CLASS` to be followed with a `;`
* Moves all `_bind_methods()` to the private scope


Naros:

PR #430, opened 2024-06-22
GH-423 Avoid reuse of default value instances

### LINKED ISSUE #423: Script default value instances are reused, causing unexpected behavior
### Describe the bug

Given the following nodes:

image

When the `arr` input pin is connected with a `Make Array` node, each call to the node will use a new `Array` instance.  For this use case that was the desired outcome.  However, if a `Make Array` node was not connected, this node used an `Array` instance from the default value cache, which was re-used on each call.  This re-use in this context causes the array to continuously grow due to the function's logic appending elements to this array on each call.

This raises the question, should default values always create a new instance.

### Expected behavior

The user behavior should be the same regardless of whether a `Make Array` node is or is not connected.

### Actual behavior

The same `Array` instance is re-used when the `arr` pin has no connection.

### How to Reproduce?

_No response_

### Godot full version

N/A

### Orchestrator version

2.0.1.stable

### Additional information

_No response_

Naros:
Fixes #423

### DIFF
--- a/src/script/instances/node_instance.cpp
+++ b/src/script/instances/node_instance.cpp
@@ -27,3 +27,28 @@ Ref<OScriptNode> OScriptNodeInstance::get_base_node()
 {
     return { _base };
 }
+
+OScriptNodeInstance::~OScriptNodeInstance()
+{
+    if (input_pin_count > 0)
+    {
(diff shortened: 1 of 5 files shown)


Naros:

PR #432, opened 2024-06-22
GH-431 Upgrade of For w/Break nodes autowires aborted

### LINKED ISSUE #431: Upgrade of For w/break should auto-wire aborted
### Describe the bug

When upgrading `For` nodes with the `Aborted` pin, the `Aborted` pin remains unwired.  This creates a change in behavior for any existing Orchestrations.  The upgrade should automatically wire `Aborted` to the same target as `Completed`, leaving it up to the user to modify the Orchestration if needed.

### Expected behavior

No change in behavior from pre `Aborted` pin.

### Actual behavior

Break triggers the `Aborted` pin and the Orchestration does not run as it previously did.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #431

### DIFF
--- a/src/script/nodes/flow_control/for.cpp
+++ b/src/script/nodes/flow_control/for.cpp
@@ -16,6 +16,7 @@
 //
 #include "for.h"
 
+#include "common/callable_lambda.h"
 #include "common/dictionary_utils.h"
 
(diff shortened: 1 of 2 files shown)


Naros:

PR #437, opened 2024-06-24
Refactor main layout & viewports

Naros:


Naros:
This needs to be updated with logic from #439

Naros:

PR #440, opened 2024-06-24
GH-428 Fix function return types that are enum/flags

### LINKED ISSUE #428: Get Keycode does not return an int
### Describe the bug

I wonder what is going on there... in the "All Actions" Dialog it says correctly it returns an int (which comes from the `Key` enum on the `keycode` Property of `InputKeyEvent` right ?)  but in the visual node (`Get Keycode`) it returns a generic Object (see screenshot) ? Is that a bug or just not implemented yet because there is an enum ?

Screenshot_2024-06-20_172258
Screenshot_2024-06-20_173645

### Expected behavior

The out pin of `Get Keycode` should be an `int`

### Actual behavior

The out pin of `Get Keycode` is a generic Object

### How to Reproduce?

See in the Screenshot

### Godot full version

4.2.2

### Orchestrator version

2.1dev3

### Additional information

_No response_

Naros:
Fixes #428

### DIFF
--- a/src/script/nodes/functions/call_function.cpp
+++ b/src/script/nodes/functions/call_function.cpp
@@ -311,24 +311,9 @@ void OScriptNodeCallFunction::_create_pins_for_method(const MethodInfo& p_method
     size_t default_index = 0;
     for (const PropertyInfo& pi : p_method.arguments)
     {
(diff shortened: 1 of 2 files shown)


Naros:

PR #445, opened 2024-06-28
GH-444 Improve pin type-safety

### LINKED ISSUE #444: Base `OScriptNodePin` on `PropertyInfo`
### Description

Godot encodes quite a bit of information in its `ProperyInfo` struct, which describes a property on a class or a method argument or return value.  If we transition `OScriptNodePin` to be based on this structure, this has a number of benefits such as making the integration of function calls much more seamless.

### Implementation ideas

Change `OScriptNodePin` class members to look something like this:
```cpp
PropertyInfo _property;
EPinType _pin_type;
EPinDirection _direction;
Variant _default_value;
Variant _generated_default_value;
BitField<Flags> _flags;
String _user_friendly_name;
String _member_parent_class; 
```
Many properties like `_pin_name`, `_type`, `_target_class`, as well as some of the flag bits would get encoded into the new PropertyInfo member named `_property`.  Additionally, this would begin to serialize hint and usage flags along with the hint string, which allows the `_file_types` variable to become persistent.

In some instances, it may be useful for a pin to reference to a member in a specific target class.  I'd prefer we consider folding part of the intent of `_target_class` into the PropertyInfo's `class_name` attribute; which keeping the other potential intent to describe the class that _owns_ the property to be enabled into the new attribute called `_member_parent_class` or perhaps `_owner_class` or something similar.

In addition, all `create_pin` functions should be changed in `OScriptNode`. They should prefer to encapsulate as much behavior as possible, ideally having the node define whether the pin is (a) execution/data and (b) the property info associated with the pin; with an optional default value.

Naros:
Fixes #444

### DIFF
--- a/cmake/extension_db.cpp.in
+++ b/cmake/extension_db.cpp.in
@@ -135,6 +135,15 @@ namespace godot
         return ExtensionDB::_singleton->_global_enums[p_name];
     }
 
+    EnumInfo ExtensionDB::get_global_enum_by_value(const StringName& p_name)
+    {
+        for (const KeyValue<StringName, EnumInfo>& E : ExtensionDB::_singleton->_global_enums)
+            for (const EnumValue& ev : E.value.values)
+                if (ev.name.match(p_name))
+                    return E.value;
+        return {};
+    }
+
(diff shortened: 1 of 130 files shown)


Naros:

PR #449, opened 2024-06-28
Updated Linux dependencies and build workflow for riscv64.

fire:
Working on a proof of concept of using librisc linux as a cross platform compiled platform for gdextensions on many platforms including windows, macos and linux.

https://github.com/fwsGonzo/libriscv/tree/master

### DIFF
--- a/.github/actions/linux-deps/action.yml
+++ b/.github/actions/linux-deps/action.yml
@@ -9,11 +9,7 @@ runs:
       run: |
         sudo apt-get update
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test
-        sudo apt-get install gcc-13 g++-13 -y
-        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-13 100
-        sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-13 100
-        sudo update-alternatives --config gcc
-        sudo update-alternatives --config g++
+        sudo apt-get install gcc g++ gcc-riscv64-linux-gnu g++-riscv64-linux-gnu -y
 
     - name: Install Ninja (Linux)
       shell: sh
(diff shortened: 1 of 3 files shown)


.github/actions/linux-deps/action.yml
```diff
@@ -9,11 +9,7 @@ runs:
       run: |
         sudo apt-get update
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test
-        sudo apt-get install gcc-13 g++-13 -y
```
fire, line 12: I wasn't able to find a gcc-13 with riscv. 

I think the riscv on ubuntu-20.04 is gcc-9.

Naros: Hi @fire, is it necessary to downgrade this to gcc-9?

fire: I think I was trying to upgrade to 24.04 which has gcc 13, but I messed up.

Naros: The main issue with using Ubuntu 24.04 is the GLIBC versions are different and many folks on certain distributions like Kubuntu for example are unable to run the compiled shared library because it uses a newer GLIBC version.

fire: The issue is trying to get a gcc-13 riscv cross compiler, not sure how to find one that isn't like 24.04.

fire: I can try it.

fire: @Naros as you requested. https://github.com/V-Sekai-fire/godot-orchestrator/actions/runs/9923424781/job/27413554767

fire: Feel free to edit my branch.

PR #450, opened 2024-06-28
Allow compiling orchestrator as doubles.

### LINKED ISSUE #447: Allow orchestrator to work on precision=double
### Description

Allow orchestrator to work on precision=double.

### Implementation ideas

See https://github.com/godot-jolt/godot-jolt/issues/355

fire:
Fixes #447 - Currently hard set to doubles for testing, needs a better design for how to release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,183 @@
+name: 🛠️ Build All
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref_name }}"
+  cancel-in-progress: true
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+#          - identifier: linux-debug
+#            name: 🐧 Linux (Debug)
+#            build_type: Debug
+#            runner: ubuntu-20.04
+#            target: template_debug
+#            platform: linux
+#            arch: x86_64
+          - identifier: linux-release
+            name: 🐧 Linux (Release)
+            build_type: Release
+            runner: ubuntu-20.04
+            target: template_release
+            platform: linux
+            arch: x86_64
+            precision: double
+#          - identifier: windows-debug
+#            name: 🪟 Windows (Debug)
+#            build_type: Debug
+#            runner: windows-latest
+#            target: template_debug
+#            platform: windows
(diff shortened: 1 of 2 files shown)


fire:
@Naros Would you know how to get more debug symbols? The unsymbolated crash makes it hard to debug.

Naros:
Can you roll the branch forward to the lastest `main` and see if that fixes it @fire ?

fire:
Made an issue https://github.com/godotengine/godot-cpp/issues/1512.

Also, I noticed that execution of the game doesn't crash.

Naros:
Ya the execution/runtime code doesn't use `Curve2D`, it's only used for rendering the `GraphEdit` connection lines, which is an EditorPlugin only code path, so that makes complete sense.  

It's still bizarre why `Curve2D.add_point` causes this issue, as I haven't seen this before.   I did check the extension_api hash and compared that to what I saw in the generated files, it seems to align other than there is an alternative hash for this function call that I don't see on others 🤷 .  Hopefully David can give some more insight on the godot-cpp issue.

Naros:
@fire I wonder if it would make sense to also do what is suggested here:
https://github.com/godotengine/godot-cpp/pull/1145

I like the idea of using `GODOT_` prefixes for godot specific bits to separate settings specifically from things that are Orchestrator only.  WDYT?

fire:
I think it's a good idea, but I am not stuck to it.

Naros:
If you don't mind, I'd appreciate it 👍

Naros:
Hi @fire so I was coming back to this PR to get it merged, and now I'm not so sure the issue is (at least a Godot one).  

In debugging, what I notice is that the hash passed into ClassDB::get_method_with_compatibility is `4175465202`; however, the value that Godot expects is `3343370600`.  When I check `gdextension_compat_hashes` for `Curve2D`, I see this:
```cpp
	mappings.insert("Curve2D", {
#ifdef REAL_T_IS_DOUBLE
		{ "add_point", 529706502, 3343370600 },
#else
		{ "add_point", 2437345566, 4175465202 },
#endif
	});
```
Given that when `REAL_T_IS_DOUBLE` isn't set in the engine, the value that we're passing into the `ClassDB` method matches, so this would lead me to think that `godot-cpp` is either (a) not handling doubles or (b) we aren't setting this in our build correctly.

Would you agree with this assessment?

PR #456, opened 2024-06-29
GH-455 Fix collapse/expand fault in all actions menu

### LINKED ISSUE #455: Certain mouse interactions cause Godot to crash in All Actions Menu
### Describe the bug

In the all actions menu perform the following steps:

1. Search for `Self`
2. Click on Variables to collapse & expand the tree
3. Click anywhere in the area of the tree widget below `Get self`
4. Editor crashes

The issue is that the dialog raises a `nothing_selected` signal in this case and the code is designed to operate on the selected item.  However, Godot returns a `TreeItem` from `Tree::get_selected` despite the Variables item being non-selectable, and so this is what caused the segfault with the code path.

Possible fixes include:
1. Remove the `nothing_selected` signal handler
2. Guard the `item_activated` callback to check that the first column is selectable before dispatch

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2, 4.3

### Orchestrator version

2.1, 2.0

### Additional information

_No response_

Naros:
Fixes #455


Naros:

PR #461, opened 2024-06-30
GH-460 Avoid variable panel exponential updates

### LINKED ISSUE #460: Variable panel update fired N times exponentially
### Describe the bug

The variable panel subscribes a callback to each Variable instance to be notified when the variable changes.  This callback was to be attached to a variable the first time the panel renders and only for new variables added thereafter; however, the callable equality check failed, resulting in the callback being subscribed N times, i.e. where N is each update, causing the panel to update N times per update, which grew exponentially. 

### Expected behavior

The variable panel should only be updated once per variable change.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2.2-stable

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #460

### DIFF
--- a/src/editor/component_panels/variables_panel.cpp
+++ b/src/editor/component_panels/variables_panel.cpp
@@ -17,6 +17,7 @@
 #include "editor/component_panels/variables_panel.h"
 
 #include "common/callable_lambda.h"
+#include "common/macros.h"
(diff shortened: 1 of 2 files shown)


Naros:

PR #462, opened 2024-06-30
Add a boolean setting to show the 'arrange' button in the UI.

fire:
I am trying to restore auto-arrange. Do you remember why it was removed?

https://github.com/Vahera/godot-orchestrator/assets/32321/68c54a55-b2e0-4042-9144-f0d063e86d99

I noticed that auto-arrange undo and redo is bugged - Will fill a bug report.

Also frames are not interacting.

I might be able to get these changed patched upstream.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -150,6 +150,7 @@ void OrchestratorSettings::_register_settings()
     _settings.emplace_back(BOOL_SETTING("ui/graph/disconnect_control_flow_when_dragged", true));
     _settings.emplace_back(BOOL_SETTING("ui/graph/show_autowire_selection_dialog", true));
     _settings.emplace_back(BOOL_SETTING("ui/graph/show_minimap", false));
+    _settings.emplace_back(BOOL_SETTING("ui/graph/show_arrange_button", false));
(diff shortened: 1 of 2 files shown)


fire:
I propose that this is kept in the settings menu off by default so I can debug it.

Naros:
Hi @fire, thanks for the contribution.

If I recall, there was an issue with auto-arrange where it places the nodes in really obscure ways stacked in really tall vertical way rather than horizontal like you would normally expect, so we disabled it.  I certainly don't mind using a toggle for this.

As for the comment nodes, those aren't `GraphFrame` types.  The `GraphFrame` class wasn't added to Godot 4.2, so we had to roll our own implementation based on `GraphNode`.  There is a GH issue to port comments to `GraphFrame` for Orchestrator 2.1, but this won't be possible for 2.0 since that must remain based on Godot 4.2 -- unless upstream backports `GraphFrame` to 4.2 ofc.

Naros:
Here's a great example, this is the event graph for `camera.torch`:

image

Now, when using the auto-arrange feature:

image

I haven't looked at the logic in `GraphEdit` in the upstream, but that really doesn't make it look any better imo.

Naros:
Applied to main (=2.1) and branch 2.0.  Congrats on your first contribution to the repository @fire 🎉

PR #466, opened 2024-06-30
GH-465 Apply "Refresh Nodes" option to all selected nodes

### LINKED ISSUE #465: Refresh nodes does not apply to selected nodes
### Describe the bug

When using the node context menu, selecting "Refresh Nodes", it does not apply to all selected nodes.  Instead, it only applies to the node that the context-menu was trigger on.

### Expected behavior

All selected nodes should be refreshed.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2.2-stable and 4.3.beta2

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #465


Naros:

PR #471, opened 2024-07-02
GH-470 Skip autowire dialog for math nodes

### LINKED ISSUE #470: Do not show autowire dialog for math nodes
### Describe the bug

While the autowire dialog is helpful in some contexts, it can be pretty frustrating in others, such as working with math nodes that can often take 2 of the same argument.  For workflow simplicity, we should disable the autowire dialog for math nodes.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:


Naros:
Fixed in #445

PR #479, opened 2024-07-05
GH-476 Fix error handling in virtual machine

### LINKED ISSUE #476: Debugger not triggered when input argument resolution fails
### Describe the bug

During the method handler, if a node that was being executed failed to resolve input values, the VM would write an error to the console; however, this did not trigger the debugger and instead exited the handler immediately.  This could also lead to issues in the new debugger code where the call stack method details were not being managed correctly.

### Expected behavior

When input resolution fails, the VM loop should break rather than return so that the error is handled by the debugger correctly and that the function's call stack is cleaned up correctly.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #476

### DIFF
--- a/src/script/vm/script_vm.cpp
+++ b/src/script/vm/script_vm.cpp
@@ -719,7 +719,7 @@ void OScriptVirtualMachine::_call_method_internal(const StringName& p_method, OS
                 context.get_error_reason()));
 
             r_return.clear();
-            return;
+            break;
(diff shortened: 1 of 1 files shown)


Naros:

PR #485, opened 2024-07-06
GH-484 Show enum, bitfield, class icons in variable panel

### LINKED ISSUE #484: Should class icons in variable panel
### Description

When a variable is defined as an enum, bitfield, or a class type, its icon should reflect those similar to basic types.

### Implementation ideas

_No response_

Naros:
Fixes #484


Naros:

PR #493, opened 2024-07-09
GH-492 Correctly resolve signal argument types

### LINKED ISSUE #492: Signal add callback does not decode arguments correctly
### Describe the bug

When adding a new signal handler, arguments are not decoded properly and are always listed as Variant types, when they should actually be potentially typed to specific native classes such as `Area3D` or `Node2D`.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #492 

This change aims to improve the creation of signal slot function handlers by ensuring that argument types are resolved as appropriate class types rather than always treating them as explicit `Variant` or basic types. In addition, this fix includes a few enhancements.

### Add signals to not yet opened Orchestrations

In prior releases, the `Orchestration` must be opened in the Orchestrator tab before a signal slot function can be added to the script.  If this was not the case, the signal would get registered in the scene but no function would be created in the script to handle the signal when fired.  

The callback handler method has been relocated to the main editor panel widget, which is always active in the Editor.  This means that if you create a signal slot handler function and pick a node that has an Orchestration on it, the callback will be handled by Orchestrator even if the script is not yet opened.  If the script isn't opened, it will be opened.  If the Orchestrator tab is not the current main focus in the Editor, it will be made active as well.  This should help with overall workflow.

### Disconnecting slot functions

A slot function, one which has a green icon to the right of its name, can now be disconnected from the signal by simply right clicking the Graph Function or User-defined Function in the component view and selecting `Disconnect`.

### DIFF
--- a/src/editor/component_panels/functions_panel.cpp
+++ b/src/editor/component_panels/functions_panel.cpp
@@ -38,6 +38,35 @@ void OrchestratorScriptFunctionsComponentPanel::_show_function_graph(TreeItem* p
     _tree->deselect_all();
 }
 
+void OrchestratorScriptFunctionsComponentPanel::_disconnect_slot(TreeItem* p_item)
+{
+    const Ref<OScript> script = _orchestration->get_self();
+    const Vector<Node*> nodes = SceneUtils::find_all_nodes_for_script_in_edited_scene(script);
+
+    const String method_name = _get_tree_item_name(p_item);
+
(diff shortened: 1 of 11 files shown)


Naros:

PR #499, opened 2024-07-10
GH-487 Changing function argument type resets default value

### LINKED ISSUE #487: Changing function argument types sets unexpected default values
### Describe the bug

When a user-defined script function is dragged onto the graph canvas and the function argument types are modified, the default value for the pin will sometimes be rendered with `<null>` for numeric/string data types.  

### Expected behavior

For user-defined functions, default values should always default to the initial specific Variant subtype value.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #487


Naros:

PR #504, opened 2024-07-11
GH-490 Make PopupMenu instances dynamic for Node/Pins

### LINKED ISSUE #490: Reduce number of PopupMenu instances in Graphs
### Description

In an Orchestration Graph, each `GraphNode` and each `GraphNodePin` generates a `PopupMenu` instance for the context-menus.  While this doesn't present an issue, it's inefficient and a single `PopupMenu` could be shared across these use cases, or at the very least a single instance spawned & freed when no longer needed.

### Implementation ideas

_No response_

Naros:
Fixes #490


Naros:

PR #507, opened 2024-07-12
GH-491 Improve graph node left/right column rendering

### LINKED ISSUE #491: Avoid left/right columns condensing into one another.
### Description

Presently a `GraphNode` can be rendered as follows:

image

Ideally this should be rendered as:

image

This avoids the left & right sides being intermingled, and it appears there is a clear division between input/output.

### Implementation ideas

_No response_

Naros:
Fixes #491


Naros:

PR #508, opened 2024-07-12
GH-496 Track editor graph state in editor cache

### LINKED ISSUE #496: Restore all open graphs on editor restart
### Description

While Orchestrator tracks what Orchestration scripts are open when the editor is closed & reopened, it does not track what are the currently opened sub-graphs within an Orchestration, such as a user-defined function.  Ideally Orchestrator should restore all the same state so that the editor takes the user back to the same place it was at before the restart.

### Implementation ideas

_No response_

Naros:
Fixes #496


Naros:

PR #509, opened 2024-07-12
[DO NOT MERGE] Test MacOS min 10.14 support

Naros:


Naros:
@NullMember, we have a user who is using MacOS 10.14 and the build based on `macos-latest` would not load and they were not seeing any messages in Godot about why the library would not start.  

I found some information on setting the macOS compatibility for 10.14, which I have tried in this PR; however, when the user attempts to use the `.dylib` file, Godot now crashes.  So while its definitely loading the file, something else is awry.  

Do you have any suggestions or ideas?

NullMember:
I’ve checked the artifact generated by the CI, and it seems the minimum OS version of the dylib is 11.0.
```
Load command 9
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 11.0
      sdk 14.0
   ntools 1
     tool 3
  version 1015.7
```

I don't have any suggestions for now but I can try to build it for 10.14 (or for 10.12, which is the minimum for Godot itself) locally. I will keep you informed of my progress.

Naros:
> I don't have any suggestions for now but I can try to build it for 10.14 (or for 10.12, which is the minimum for Godot itself) locally. I will keep you informed of my progress.

My guess is maybe `macos-latest` which is macOS 14 can't go back to 10.x at this point?

NullMember:
> > I don't have any suggestions for now but I can try to build it for 10.14 (or for 10.12, which is the minimum for Godot itself) locally. I will keep you informed of my progress.
> 
> My guess is maybe `macos-latest` which is macOS 14 can't go back to 10.x at this point? Perhaps we bump the runner back to macos 12?

My machine (M1) is running macOS 14, and I can't compile for a minimum OS version before 11. ARM64 support was first added in macOS 11. This problem could be related to the x86_64;arm64 universal binary. We could try compiling the ARM64 and x86_64 versions separately using different runners.

Naros:
My only concern is it does not look like Godot differentiates between arm64 and x86_64, see:
https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_cpp_example.html#using-the-gdextension-module

So we wouldn't be able to distribute a single artifact any longer, which would be problematic for the asset library.

Naros:
So we'd need to test by updating the `orchestrator.gdextension` file and see whether or not `macos.release.arm64` and `macos.release.x86_64` are used for the different MacOS architectures by Godot.  If they are, then we can try your suggestion.

NullMember:
Latest artefact shows minos version 10.14

```
Load command 8
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 10.14
      sdk 14.0
   ntools 1
     tool 3
  version 1015.7
```

Naros:
That's good, unfortunately the user says this still gives them a crash with Godot 4.2.2 🤔 
Any other ideas @NullMember ?

NullMember:
> That's good, unfortunately the user says this still gives them a crash with Godot 4.2.2 🤔 Any other ideas @NullMember ?

I think we need more detailed report on why Godot was crashing. Godot can be started from terminal using verbose option. It could give us an idea how to solve the problem

Naros:
The user shared these output screenshots:
image
image

Does this help @NullMember ?

Naros:
We can reopen if we get more action on this.

PR #513, opened 2024-07-13
GH-498 Improve NodePath dialogs for nodes and properties

### LINKED ISSUE #498: Tween property function call does not allow selecting properties from list
### Describe the bug

When creating a property tweener, the `Tween Property` function call node should provide the user with a list of properties based on the object that is connected to the object pin.  Additionally, if the object pin is later disconnected, the value of the property pin should be to its default.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #498

### DIFF
--- a/src/common/property_utils.cpp
+++ b/src/common/property_utils.cpp
@@ -112,6 +112,17 @@ namespace PropertyUtils
         return Variant::get_type_name(p_property.type);
     }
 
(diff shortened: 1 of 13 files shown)


Naros:

PR #514, opened 2024-07-13
GH-512 Correctly resolve target object

### LINKED ISSUE #512: Function target type not always resolved
### Describe the bug

Given a nested instantiated scene as follows:
* Main
  * Widget (nested)
    * Control
      * Sprite2D

When dragging from a `Get Scene Node` that refers to `Control/Sprite2D`, shown below, if the current edited scene is the outer scene that has `Main` as the root, then the `Get Texture` function node is added correctly; however, when working within the nested scene, as shown here, the `Get Texture` function is added incorrectly:

image

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #512


Naros:

PR #515, opened 2024-07-13
GH-511 Add label to Enum/Bitfield return function values

### LINKED ISSUE #511: Output enum/bitfield pins should be labeled
### Describe the bug

Currently enum/bitfield output pins are treated like normal basic types such as Strings, Integers, and Floats where only the icon is shown.  The enum/bitfield class name should be shown to give more context without relying on the tooltip.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #511


Naros:

PR #519, opened 2024-07-13
GH-473 Improve view documentation jump for nodes

### LINKED ISSUE #473: Goto help fails for CallMemberFunctionNode
### Describe the bug

When using the "View Documentation" option on a CallMemberFunctionNode where the target is the "Get Scene Tree" node, it fails to jump to the function, in this case the `SceneTree::quit()` function.

There are other use cases such as "Create Tween", "Tween Property", etc.

It appears the `_reference.target_class_name` is not being serialized correctly.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.3.beta2

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #473

### DIFF
--- a/src/script/nodes/data/compose.cpp
+++ b/src/script/nodes/data/compose.cpp
@@ -384,6 +384,15 @@ String OScriptNodeComposeFrom::get_icon() const
     return "Instance";
 }
 
+String OScriptNodeComposeFrom::get_help_topic() const
+{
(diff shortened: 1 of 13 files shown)


Naros:
Applied to main (=2.1) only.

PR #520, opened 2024-07-13
GH-472 Show event actions when dragging from input pins

### LINKED ISSUE #472: When dragging from a float output pin, event-based calls are eligible and shouldn't
### Describe the bug

Given the following graph:

image

Dragging from any of the float outputs of X, Y, or Z shows these as options:

image

While in Godot terms, these methods do accept input arguments and so technically the logic is correct, these are rendered where the input arguments are actually outputs in an Orchestration PoV, and shouldn't be possible choices.

### Expected behavior

The `process` and `physics process` functions shouldn't be rendered.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2.2-stable and 4.3.beta2

### Orchestrator version

2.1.dev3

### Additional information

_No response_

Naros:
Fixes #472

### DIFF
--- a/src/editor/graph/graph_edit.cpp
+++ b/src/editor/graph/graph_edit.cpp
@@ -1374,6 +1374,7 @@ void OrchestratorGraphEdit::_on_attempt_connection_to_empty(const StringName& p_
         return;
 
     OrchestratorGraphActionFilter filter;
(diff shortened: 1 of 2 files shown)


Naros:

PR #530, opened 2024-07-15
GH-529 Improve add component item workflow

### LINKED ISSUE #529: Improve component panel workflow
### Description

In the component panel, it is not clear why one would or how you would add new Godot event nodes to any other graph beyond the `EventGraph`.  

When creating override event functions, the workflow should first target the current open graph. If the open graph is an event graph, it is permissible to override the event function. This will allow users to create new event graphs and add new event node overrides. If the open graph is a function graph, the workflow will fall back to adding the event override node to the main event graph.

In addition, when adding new elements to the component panel (graphs, functions, variables, and signals), users must manually activate the newly added item to view the event graph, function graph, variable, or signal details.  The workflow should be updated so that when a new item is added, it becomes editable, but the item-activated state should be triggered before the edit.  

For example, when adding a new event graph, the graph is added to the Orchestration, the panel is updated, the new element is activated, triggering the opening of the graph, and the item in the component panel is then selected and edited to allow the user to rename the newly added graph. The same workflow should occur for functions, variables, and signals.

### Implementation ideas

_No response_

Naros:
Fixes #529

### DIFF
--- a/src/editor/component_panels/component_panel.cpp
+++ b/src/editor/component_panels/component_panel.cpp
@@ -51,7 +51,7 @@ void OrchestratorScriptComponentPanel::_tree_add_item()
     if (_handle_add_new_item(new_name))
     {
         update();
-        _find_child_and_activate(new_name);
+        _find_child_and_activate(new_name, true, true);
     }
 }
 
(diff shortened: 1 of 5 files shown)


Naros:

PR #536, opened 2024-07-15
Add Mac, Linux, and Web ignores to .gitignore file

aaronfranke:
This is a small PR that adds macOS and iOS `.dylib`, Linux and Android `.so`, and Web `.wasm` ignores. This prevents build artifacts from non-Windows platforms from showing up in diffs.


Naros (approved):

Naros:
Applied to main (=2.1) and 2.0.

PR #540, opened 2024-07-17
GH-539 Fix placeholder instance clean-up

### LINKED ISSUE #539: Extension ScriptInstance references not cleared in placeholder/instances array
### Describe the bug

Each instance of an Orchestration maintains an internal list of references to its instances.  For runtime, these are ScriptInstance objects and for non-tool editor instances, these are called PlaceholderScriptInstance objects.

On the engine side, not only are these instances created by calls to the `Script` contract, but there are also methods that the engine calls to clean up these references on the `Script` contract, such as `_placeholder_erase`.  The problem is that these methods are not called by the `ScriptInstanceExtension`.

For Godot 4.3 and 4.2, we need to account for this in our code, so that when the `ScriptInstanceExtension` calls its delete function, that we appropriately clean up all references in the `Script` instance for its runtime and/or placeholder instances.

For Godot 4.4, we should try and align this behavior with the `ScriptExtension` contract such that if the instance is a placeholder, the `ScriptInstanceExtension` will do the following during its destructor:

```cpp
	virtual ~ScriptInstanceExtension() {
		if (native_info->is_placeholder()) {
			get_script()->placeholder_erase(instance);        
		}
		if (native_info->free_func) {
			native_info->free_func(instance);
		}
		...
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.2, 4.3

### Orchestrator version

2.1.dev4, 2.0.2.rc2

### Additional information

_No response_

Naros:
Fixes #539

### DIFF
--- a/src/script/instances/instance_base.cpp
+++ b/src/script/instances/instance_base.cpp
@@ -92,6 +92,11 @@ void OScriptInstanceBase::init_instance(OScriptInstanceInfo& p_info)
     p_info.get_language_func = [](void* p_self) {
         return ((OScriptInstanceBase*)p_self)->get_language()->_owner;
     };
+
+    p_info.is_placeholder_func = [](void* p_self) -> GDExtensionBool {
+        return ((OScriptInstanceBase*) p_self)->is_placeholder();
+    };
+
 }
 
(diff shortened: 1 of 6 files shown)


Naros:

PR #544, opened 2024-07-20
GH-534 Correctly resolve self for NodePath popups

### LINKED ISSUE #534: The `Get Self` node does not allow selecting property path for `Tween Property`.
### Describe the bug

When using a `Get Self` node as the object for `Tween Property` function call, the property dialog is not shown.  However, replacing the `Get Self` with a `Get Scene Node` of the same node reference works.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.dev4

### Additional information

_No response_

Naros:
Fixes #534


Naros:

PR #545, opened 2024-07-20
GH-532 Align Get Scene Node class on path change

### LINKED ISSUE #532: Changing node path on Get Scene Node does not update pin class type
### Describe the bug

Start with a simple node that points to a `CharacterBody3D`:

image

Change the node path to point to a different node, such as a node for a `MeshInstance3D`:

image

The pin class type does not change.

### Expected behavior

Pin class type should change.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.3.beta3

### Orchestrator version

2.1.dev4

### Additional information

_No response_

Naros:
Fixes #532


Naros:

PR #546, opened 2024-07-20
GH-338 Support variables as script global class types

### LINKED ISSUE #338: GDScript named-class scripts don't appear in variable type selection
### Describe the bug

When a custom type is defined in GDScript using `class_name`, it is not shown in the variable type selection.

### Expected behavior

Custom type should be selectable in the variable type dialog.

### Actual behavior

It's not present.

### How to Reproduce?

_No response_

### Godot full version

4.2.1-stable

### Orchestrator version

2.1.dev1

### Additional information

So these global class names are provided via `ProjectSettings::get_singleton()->get_global_class_list()`.  

The main issue here is that there isn't a way that's clear for Orchestrator to obtain details about the custom types properties and methods without constructing an instance of the `GDScript` first, and then getting the data.  We need to investigate if there is a way to get this from the language or some metadata cache in Godot to avoid this extra allocation.

Naros:
Fixes #338


Naros:

PR #552, opened 2024-07-21
GH-497 Function/Signals support all class/argument types

### LINKED ISSUE #497: Allow function/signals to define arguments of specific class types
### Description

Currently if a user wants to pass a specific class type into a function, they need to pass the object using `Object` or `Any` (variant), which requires that the user recast the object to its desired type inside the function.  This is unnecessary and function and signal arguments should be allowed to define an explicit class type or enumeration they represent, just like variables.

### Implementation ideas

_No response_

Naros:
Fixes #497


Naros:

PR #558, opened 2024-07-21
GH-556 Generalize "select type" dialog

### LINKED ISSUE #556: PropertyContainer inspector widget has several inconsistencies
### Describe the bug

1. If the property has the type `Variant,` opening the type dialog does not place the user on the `Any` type but some obscure enum type that scores highest to the term `Variant.`  In this use case, it should auto-select `Any`.
2. Once the dialog has been opened, selecting `Cancel` causes Godot to crash.
3. The dialog title always reads `Select Variable Type`, this should either be generalized or specific.
4. The dialog's recent history and favorites map to the variable classification; this should be unique to function/signals.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #556


Naros:

PR #564, opened 2024-07-21
GH-563 Relax validation on function result node

### LINKED ISSUE #563: Function return node expects a connection, even when setting a default return value
### Describe the bug

When validating the Function return node, the validation still expects a connection if the return type is a `Boolean` or some other basic type where a default value can be supplied.

### Expected behavior

Validation should not require a connection if a default value can be supplied.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #563


Naros:

PR #569, opened 2024-07-22
GH-568 Relax property validation

### LINKED ISSUE #568: Node validation too strict for Property Get/Set nodes when target cannot be resolved
### Describe the bug

If the target object cannot be resolved, the validation fails; where it may be valid.
The validation should fallback to checking native classes before failing.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc1

### Additional information

_No response_

Naros:
Fixes #568


Naros:

PR #576, opened 2024-07-23
GH-575 Correctly focus script on editor breakpoints

### LINKED ISSUE #575: When selecting stack frame, Orchestration is not opened.
### Describe the bug

When running an Orchestration in the Editor, if the Orchestration causes the debugger to create a breakpoint due to an error, the stack frame UI will show the relative point where the failure occurred:

image

The problem is that clicking on the stack frame does not open the Orchestration and focus the node if it isn't already open.  If the Orchestration is open, the specific Orchestration will be brought forward & the node focused.

_NOTE: that while the node is focused if the Orchestration is open, the file list "active script" marker is not updated._

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.3.beta3

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #575

### DIFF
--- a/src/editor/editor_panel.cpp
+++ b/src/editor/editor_panel.cpp
@@ -725,9 +725,16 @@ void OrchestratorEditorPanel::_goto_script_line(const Ref<Script>& p_script, int
                 _show_editor_viewport(file.file_name);
                 // Goto node
                 file.viewport->goto_node(p_line + 1);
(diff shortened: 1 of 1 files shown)


Naros:

PR #580, opened 2024-07-23
GH-579 Raise breakpoint when input action has no action name

### LINKED ISSUE #579: Input action should report error if action name is empty
### Describe the bug

Presently the `InputAction` node does not throw an error if the user fails to provide an action name, and instead allows the script to continue running without raising a debugger breakpoint.

### Expected behavior

A breakpoint should be raised with a sufficient error.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.3.beta3

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #579

### DIFF
--- a/src/script/nodes/input/input_action.cpp
+++ b/src/script/nodes/input/input_action.cpp
@@ -40,6 +40,12 @@ public:
             return -1 | STEP_FLAG_END;
         }
 
+        if (_action_name.is_empty())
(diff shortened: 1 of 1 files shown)


Naros:
Applied to main (=.21) and branch 2.0.

PR #586, opened 2024-07-24
GH-582 Auto-select first action when node first placed

### LINKED ISSUE #582: InputAction pin does not auto-select first element
### Describe the bug

Inside the `OrchestratorGraphNodePinInputAction::_populate_action_list()` method, if no item is selected, the pin's default value is assigned an empty variant, which can lead to a series of spammed errors about unable to locate an input action with an empty string. 

### Expected behavior

There should be some type of error or the item shown as being selected should be auto-selected.

### Actual behavior

Runtime throws an error

### How to Reproduce?

1. Add an `Input` singleton to the graph.
2. Drag away and select the `Is Action Pressed` function
3. Save and run

### Godot full version

4.3.beta3

### Orchestrator version

2.1.rc2

### Additional information

For 2.1, it would make sense for now to enforce auto-selection of the first action at index 0 when first placing the node.  This aligns with what a user would initially expect, given there is no `--Select--` type of placeholder option.

For 2.2, we should rethink this specific pin's validation routine. Ideally, it would be helpful to introduce a `--Select--` placeholder and then throw an error during validation and runtime if the user never explicitly picked an action name. The main issue here, specifically with runtime validation, is how to know what member function calls require such extended validation or, more specifically, which function calls use this input action list of values since this is handled at the editor layer and not the underlying resource layer, we're currently limited.

Naros:
Fixes #582

### DIFF
--- a/src/editor/graph/pins/graph_node_pin_input_action.cpp
+++ b/src/editor/graph/pins/graph_node_pin_input_action.cpp
@@ -62,7 +62,10 @@ void OrchestratorGraphNodePinInputAction::_populate_action_list()
     }
 
     if (!found)
-        _pin->set_default_value(Variant());
+    {
+        _button->select(0);
+        _pin->set_default_value(_button->get_item_text(0));
+    }
 }
 
 Control* OrchestratorGraphNodePinInputAction::_get_default_value_widget()


Naros:

PR #587, opened 2024-07-24
GH-583 Always show execution pins on user-defined script functions

### LINKED ISSUE #583: Custom user-function that starts with `get_`, has no arguments, and returns a value does not render correctly
### Describe the bug

When defining a custom user function that starts with `get_` the interface assumes there are no execution pins, and this is incorrect for user-defined functions and should only apply to Godot-specific function calls.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #583


Naros:

PR #597, opened 2024-07-27
GH-595 Improve initial rendering of variable type selection

### LINKED ISSUE #595: Variable type selection dialog can be overwhelming
### Description

When the variable type dialog is first opened, the number of options can seem overwhelming.  To address this, a suggestion was made to do the following:

- Changing the variable type won't populate the search box, it's always opened as empty.  This means that no item will be initially selected.
- All top-level tree items should be collapsed initially (basically only allowing basic types to be visible)
- All global enumerations should be moved under `Types > Global Enumerations > [Enum Name]`.

### Implementation ideas

_No response_

Naros:
Fixes #595 

- All search items (excluding root item) are collapsed when the search box is empty.
- There is no initial search text in the search box.
- All global enums and bitfields have been relocated under new `Types/Global Enum` and `Types/Global Bitfield` categories.
- Fixes clearing search box to reset favorite button press state.

### DIFF
--- a/src/editor/inspector/property_type_button_property.cpp
+++ b/src/editor/inspector/property_type_button_property.cpp
@@ -21,7 +21,7 @@
 
 void OrchestratorEditorPropertyVariableClassification::_property_selected()
 {
-    _dialog->popup_create(true, true, get_edited_object()->get(get_edited_property()), get_edited_property());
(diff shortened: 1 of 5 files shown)


Naros:

PR #598, opened 2024-07-27
GH-596 Render `int`, `float`, and `bool` with user-friendly names

### LINKED ISSUE #596: Variable type names should be shown consistently
### Describe the bug

In the variable type selection dialog, the `int` and `bool` types use their `Variant::get_type_name` naming, which is inconsistent with the other types that are shown.  Additionally, the `float` type is not using capitalize-case like the other, too.

Notes:
_Users should be able to type `int` and the `Integer` type should match.
Users should be able to type `bool` and the `Boolean` type should match._

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #596


Naros:

PR #599, opened 2024-07-27
GH-592 Add optional delete node confirmation dialog

### LINKED ISSUE #592: Add node deletion confirmation
### Description

In an effort to avoid accidental deletion of nodes using trackpads and context menus, users should have an option to enable deletion confirmation. This will show a dialog requesting confirmation before deleting nodes.

### Implementation ideas

_No response_

Naros:
Fixes #592


Naros:

PR #600, opened 2024-07-27
GH-593 Fix collapse to function bugs

### LINKED ISSUE #593: Collapse to function looses pin type details & doesn't assign unique pin names
### Describe the bug

GIven this graph:

image

When collapsing the gold-bordered nodes to a function, this yields an unexpected outcome in the function graph:

image

The function entry node does not properly assign the pin types, and instead just defaults to `Object` rather than `Input`.  Additionally, and more importantly, the connection wires are mistakenly drawn due to this:

image

Both input pins are named `target`.  By changing the pin names from `target` to `target1` and `target2`, the connection wires are drawn correctly. 

Finally, the `Result Node` seems randomly placed and not wired to anything.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

4.3.rc1

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #593 

- Pins should retain type
- Multiple pins should not have the same name
- Optional return node
- Script function call nodes use argument namse


Naros:

PR #604, opened 2024-07-27
GH-603 Permit selecting abstract types for variable definitions

### LINKED ISSUE #603: Variables cannot be abstract types
### Describe the bug

A variable is simply a declaration, and it shouldn't care whether it's defined with a concrete, instantiable class type or an abstract type that cannot be instantiated independently. It's meant to provide storage for a value, nothing more.

Given the following GDScript:
```python
var _my_resource : Resource

func _ready() -> void:
  _my_resource = ResourceLoader.load("res://my_resource.tres")
```

This is quite a trivial and unlikely example, but it shows that regardless of the derived resource type, the user can store the loaded `Ref<Resource>` into a `Resource` variable type.  Now, one could use `Variant` in this context by simply not specifying a type on the `_my_resource` declaration and instead choosing to cast the variable later when used.  These are all valid use cases.

The problem right now is that Orchestrator doesn't permit the former; you either choose the `Any` (Variant) storage approach described above or pick a concrete type, enforcing the casting at storage type, not at read time.  Orchestrator shouldn't be opinionated and should allow the user to select whichever strategy best fits their use case.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #603

### DIFF
--- a/src/editor/inspector/property_type_button_property.cpp
+++ b/src/editor/inspector/property_type_button_property.cpp
@@ -93,6 +93,7 @@ void OrchestratorEditorPropertyVariableClassification::_notification(int p_what)
         _dialog->set_popup_title("Select variable type");
         _dialog->set_data_suffix("variable_type");
         _dialog->set_base_type(_base_type);
+        _dialog->set_allow_abstract_types(true);
         add_child(_dialog);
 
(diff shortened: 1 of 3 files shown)


Naros:

PR #606, opened 2024-07-28
GH-605 Improve numeric default value widget validation

### LINKED ISSUE #605: Numeric default value field does not coerce values
### Describe the bug

When a numeric pin (`int` or `float`) accepts a value that does not conform to the `int` or `float` data type, while the value in the default value property is coerced, the UI is not properly coerced.  For example, entering a value of `abc` should result in the field being set to `0`.  If the data type is `int` and entering a value of `25.25` should result in the field being set to `25`.

### Expected behavior

1. Entering a value of `abc` should set the `LineEdit` value to `0`.
2. Entering a value of `25.25` should set the `LineEdit` value to `25` when the type is `Int`.

### Actual behavior

The entered value is just retained without any field updates.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #605

### DIFF
--- a/src/editor/graph/pins/graph_node_pin_numeric.cpp
+++ b/src/editor/graph/pins/graph_node_pin_numeric.cpp
@@ -16,55 +16,65 @@
 //
 #include "graph_node_pin_numeric.h"
 
-#include <godot_cpp/classes/line_edit.hpp>
-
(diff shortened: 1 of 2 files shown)


Naros:

PR #608, opened 2024-07-28
GH-607 Improve changing variable types

### LINKED ISSUE #607: Changing variable type resets default values
### Describe the bug

When changing a variable type, the `Variable` default value is always reset to its initial value.  Additionally, if there are any `VariableSet` nodes in the graph, their default values are reset if they have no connections and have been manually changed.

### Expected behavior

When changing a variable's type, if conversion can be done between types, then the variable's default value should be converted.  Additionally, the act of changing a variable's type should not reset the manually assigned default values on `VariableSet` nodes unless the value cannot be converted.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc2

### Additional information

_No response_

Naros:
Fixes #607 

### Change Highlights

- Add confirmation dialog with warning
- Attempt to coerce most common types between one another
- Retain VariableSet manual default values if they can be converted
- Break any connections with Variable Get/Set if types are incompatible

### DIFF
--- a/src/common/variant_utils.cpp
+++ b/src/common/variant_utils.cpp
@@ -120,4 +120,179 @@ namespace VariantUtils
 
         return UtilityFunctions::type_convert(Variant(), p_type);
     }
+
+    Variant convert(const Variant& p_value, Variant::Type p_target_type)
+    {
+        if (Variant::can_convert(p_value.get_type(), p_target_type))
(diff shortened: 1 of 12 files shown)


Naros:

PR #623, opened 2024-07-29
GH-622 Revert commit 8f15816b092c9a92b878db319a496430925f8816

### LINKED ISSUE #622: Adding execution pins automatically to `is_` or `get_` prefixed script functions introduces backward incompatibility
### Describe the bug

In https://github.com/Vahera/godot-orchestrator/issues/583, a change was added that allows the `is_` and `get_` user-defined functions always to render execution pins. The problem is that this introduces a backward incompatibility with existing Orchestrations, making them inoperable.

### Expected behavior

Old orchestrations should continue to work.

### Actual behavior

Old orchestrations that have `is_` or `get_` prefixed functions cease to work.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

To make this backward compatible, we could add a flag to each `OScriptNodeCallScriptFunction` node that tracks whether to display the execution pins. By default when loading older orchestrations and the function name begins with `is_` or `get_`, the flag would be set to `false` to remain backward compatible.

For newly added/placed nodes, the flag would be set to `true` for all script function calls.

Question: Should this flag be togglable in the inspector?

Naros:
Fixes #622

### DIFF
--- a/src/script/nodes/functions/call_script_function.h
+++ b/src/script/nodes/functions/call_script_function.h
@@ -35,7 +35,6 @@ protected:
 
     //~ Begin OScriptNodeCallFunction Interface
     bool _is_method_info_serialized() const override { return false; }
-    bool _has_execution_pins(const MethodInfo& p_method) const override { return true; }
     bool _use_argument_class_name() const override { return false; }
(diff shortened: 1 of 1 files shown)


Naros:

PR #631, opened 2024-08-01
GH-630 Enhance `VariableGet` with validation mode

### LINKED ISSUE #630: Add option for `Variable Get` to use a "validation mode"
### Description

When a variable is typed as `Variant::OBJECT`, the underlying variable will initially start being `null`.  This means that users need to be mindful of this and check for this use case in their code.  However, rather than adding the need for the users to introduce branch conditions in the graph, an optimized `Variable Get` node can automatically handle this:

image

Conceptually, when the variable is `null`, the node exits the `Is Not Valid` output pin.  If the variable is not `null`, it exits the `Is Valid` pin.  This avoids the need for using `is_instance_valid` or null checks for such variables.

### Implementation ideas

_No response_

Naros:
Fixes #630


Naros:

PR #633, opened 2024-08-01
GH-632 Add new function call node color variants

### LINKED ISSUE #632: Add additional color variants for orchestration & script function calls
### Description

It would be helpful to identify different types of function calls, such as those that are Godot functions, those that are from the orchestration, and those that are part of another script elsewhere.

image

### Implementation ideas

_No response_

Naros:
Fixes #632


Naros:

PR #637, opened 2024-08-01
Improve Local Variables Initialization & Assignment UX

Naros:


Naros:
It does appear that duplicating an Assign Node does not reset its type back to `Any`.

Naros:
> It does appear that duplicating an Assign Node does not reset its type back to `Any`.

Fixed.

Naros:

PR #645, opened 2024-08-02
GH-642 Allow `Object` to connect to `Boolean` pins

### LINKED ISSUE #642: Output object pins should be connectable to Boolean input pins
### Description

In GDScript, its common to use the `if` statement to validate an object's existence, i.e.:
```python
if my_object:
  # do something here
```
By allowing an `Object` output pin type to connect to a `Boolean` input pin type, we can easily perform such tests.

### Implementation ideas

_No response_

Naros:
Fixes #642


Naros:

PR #649, opened 2024-08-03
GH-646 Function graph validation for return node links

### LINKED ISSUE #646: Return node is only executed if its explicitly connected
### Describe the bug

Given the following graph:

image

The VM will execute the `F 00` node and the `Branch` node, immediately exiting the function call. This results in the function's return value being set to `null` rather than `3`.  If the `False` output pin from the `Branch` node were connected to the `Return Node`, then the output would have been `3` as well.

### Expected behavior

Any top-level (non-nested) control flow pin should implicitly be connected to the `Return Node` if one exists and has return values.  This guarantees that the graph's execution works like that of a text-based program, i.e.:

```python
func f00:
  if false:
    for x in range(1, 10):
      # do loop stuff
  return 3
```

### Actual behavior

The `Return Node` only fires when the control flow exits the first for-loop's completed pin.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc3

### Additional information

_No response_

Naros:
Fixes #646

### DIFF
--- a/src/script/function.h
+++ b/src/script/function.h
@@ -18,6 +18,7 @@
 #define ORCHESTRATOR_SCRIPT_FUNCTION_H
 
 #include "common/guid.h"
+#include "orchestration/build_log.h"
 
 #include <godot_cpp/classes/resource.hpp>
 #include <godot_cpp/variant/dictionary.hpp>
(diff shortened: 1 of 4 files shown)


Naros:

PR #659, opened 2024-08-04
GH-658 Add version pick to plug-in updater subsystem

### LINKED ISSUE #658: Support Godot Compatibility in Version Updater
### Description

The current iteration of the plug-in's version updater needs to consider Godot compatibility details.  It simply downloads the latest build from GitHub and installs it.

### Implementation ideas

To provide the best UX possible, several fundamental changes should be considered.

### Version Picker

The bulk of this change revolves around providing users with a version picker dialog, allowing them to select which version of the available ones they can install, as shown here:

image

The second column labeled `Godot Compatibility` will indicate the minimum Godot `major.minor` version for using that build of Orchestrator. An icon indicator will be _green_ if the version is compatible and _yellow_ if it is not.  

Selecting `Download and Install` for a version that is not compatible will inform the user that the version is incompatible and that they should update their Godot Editor first.

The `Filter` option button in the top-level will allow the user to quickly filter the list of builds based on `All releases` or `Stable only` in case the user.  This is helpful if the user has enabled notifications for pre-releases but only wants to install a stable build.  

The checkbox in the top-right corner toggles a project setting that controls whether the notification subsystem will alert the designer about pre-releases or only stable releases. This checkbox allows for quick modification of this setting in the update workflow; but users can also adjust this in the `Project Settings > Orchestrator > Settings`:

image

Pressing the `Release Notes` button after selecting a version redirects the user to either GitHub's release notes for that release or to the blog post for the release, if the `release_manifests.json` has a blog URL link in the metadata.

And finally, after downloading and installing the plug-in, the user will be presented with a popup to restart the editor.

Naros:
Fixes #658


Naros:

PR #663, opened 2024-08-04
GH-662 Exclude loop-based pins from return value check

### LINKED ISSUE #662: Validation fails when loop body pins have no connections
### Describe the bug

The `For`, `For Each`, and `While` loops do not exclude their loop-body pins when unconnected from the validation check.

### Expected behavior

Loop-specific ports shouldn't require connection to return node.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #662

### DIFF
--- a/src/script/node.h
+++ b/src/script/node.h
@@ -391,6 +391,9 @@ public:
     /// @return if the node can be duplicated
     virtual bool can_duplicate() const { return true; }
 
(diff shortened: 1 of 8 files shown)


Naros:

PR #672, opened 2024-08-05
GH-648 Permit any output to connect to boolean input

### LINKED ISSUE #648: Allow connecting all basic types to a boolean input pin
### Description

Godot has a concept called `Variant::booleanize` where any Variant type can be applied in a boolean operator condition to determine whether or not it evaluates to true or false. This can be useful in graphs to minimize the number of nodes needed to test specific conditions.

There is a few caveats, such as types like the Vector, Projection, Quaternion, and others have what is called a baseline representation of their internal indexed/keyed components that represent a `false` condition.  For example, `Vector2`, `Vector3`, and `Vector4` all will return false when their internal x, y, z, and w fields equal `0`. Similarly when Projection, Quaternion, and others internally represent their "Identity" or default value when using `Projection()` or `Quaternion()`, this too represents `false`.

### Implementation ideas

Adjust `OScriptNodePin::can_accept` to always allow connecting to `BOOL` input pins.

Naros:
Fixes #648

### DIFF
--- a/src/script/node_pin.cpp
+++ b/src/script/node_pin.cpp
@@ -474,6 +474,10 @@ bool OScriptNodePin::can_accept(const Ref<OScriptNodePin>& p_pin) const
     if ((!is_execution() && p_pin->is_execution()) || (is_execution() && !p_pin->is_execution()))
         return false;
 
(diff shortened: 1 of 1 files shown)


Naros:

PR #676, opened 2024-08-07
GH-675 Relax validation on function arguments with default values

### LINKED ISSUE #675: Validation requires connection even for input pins that have nullable defaults
### Describe the bug

A classic example of a nullable input value is `Tree::create_item(item: TreeItem = null, int index = -1)`.  In such cases, the validation should not expect there to be a connection as this is a typical expectation to create the root `TreeItem`.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1 / 2.0

### Additional information

_No response_

Naros:
Fixes #675

### DIFF
--- a/src/common/method_utils.cpp
+++ b/src/common/method_utils.cpp
@@ -135,4 +135,8 @@ namespace MethodUtils
         return signature;
     }
 
+    size_t get_argument_count_without_defaults(const MethodInfo& p_method)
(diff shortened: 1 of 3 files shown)


Naros:

PR #677, opened 2024-08-07
GH-674 Fix Await signal to work with parameterized signals

### LINKED ISSUE #674: Await Signal node does not work with signals that emit parameters
### Describe the bug

When the `Await Signal` node is used with signals that are argumentless, the node reacts to the signal when raised; however, if the signal has arguments as part of the event being fired, the await node does not react.

### Expected behavior

The node should react to both parameterless and parameterized signals.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1 / 2.0

### Additional information

_No response_

Naros:
Fixes #674

### DIFF
--- a/src/script/vm/script_state.cpp
+++ b/src/script/vm/script_state.cpp
@@ -18,45 +18,34 @@
 
 #include "script/instances/node_instance.h"
 
-void OScriptState::_signal_callback(const Array& p_args)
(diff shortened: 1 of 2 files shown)


Naros:

PR #679, opened 2024-08-08
Update copyright headers

Naros:


Naros:
Some new files were added recently, this will need to be updated.

Naros:

PR #681, opened 2024-08-09
GH-680 Fix loading external resources

### LINKED ISSUE #680: Orchestrations fail to load external resources
### Describe the bug

When a variable refers to a default value such as a `Resource` like a `Texture2D` that refers to a file, the Orchestration will no longer be loadable and will fail with this error:
```
* src\script\serialization\text_loader_instance.cpp:348 - res://tests/await/await_test.torch:3 - ext_resource, invalid UID: uid://bebpqd2 - using text path instead: res://icon.svg
* src\script\serialization\text_loader_instance.cpp:515 - res://tests/await/await_test.torch:470 - Parse Error: Unknown identifier: 'ExternalResource'.
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.1.rc4

### Additional information

_No response_

Naros:
Fixes #680

### DIFF
--- a/src/script/serialization/binary_loader_instance.cpp
+++ b/src/script/serialization/binary_loader_instance.cpp
@@ -16,11 +16,14 @@
 //
 #include "script/serialization/binary_loader_instance.h"
 
+#include "binary_saver_instance.h"
 #include "common/string_utils.h"
 #include "common/version.h"
 #include "script/script.h"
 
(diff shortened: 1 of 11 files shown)


Naros:

PR #692, opened 2024-08-10
GH-691 Use function argument names rather than class names

### LINKED ISSUE #691: Use function argument names rather than class names
### Description

Currently, the call function node will use the class name if the argument has a class type. This should be adjusted to fallback to using the function argument names instead for user usability reasons.

### Implementation ideas

_No response_

Naros:
Fixes #691


Naros:

PR #726, opened 2024-08-12
GH-722 Correctly copy node state using `ui_copy` action

### LINKED ISSUE #722: Copying Nodes with `Chain` enabled causes crashes
### Describe the bug

Using `Chain` flag does not get copied which results in unexpected behavior connections. When those connections get touched Godot crashes.

### Expected behavior

`Chain` flag should be copied - i believe all Node properties should be. If not the connection should not be made.

### Actual behavior

`Chain` flag is disabled but a connection for it is created.

### How to Reproduce?

1. Create a new EventGraph
2. Add a `AudioServer` singleton
3. Add method `Get Bus Channels`
4. Duplicate `Get Bus Channels`
5. On first `Get Bus Channels` enable the `Chain` flag
6. Connect Event Connectors between the two `Get Bus Channels` Nodes
7. Connect Target Connectors between the two `Get Bus Channels` Nodes
image
8. Select all Nodes and press `Ctrl+c` to copy the Selection
9. Press `Ctrl+p` to paste the copied Nodes >> See the `Chain` flag disabled and a weird connection between the two `Get Bus Channels` Nodes
image
10. Trying the remove this odd connection causes Godot to crash

### Godot full version

4.3.rc3

### Orchestrator version

2.1.rc4

### Additional information

_No response_

Naros:
Fixes #722

### DIFF
--- a/src/editor/graph/graph_edit.cpp
+++ b/src/editor/graph/graph_edit.cpp
@@ -1851,8 +1851,7 @@ void OrchestratorGraphEdit::_on_copy_nodes_request()
 
             const int id = script_node->get_id();
             _clipboard->positions[id] = script_node->get_position();
-            // Be sure to clone to create new copies of pins
-            _clipboard->nodes[id] = script_node->duplicate(true);
+            _clipboard->nodes[id] = _script_graph->copy_node(id, true);
         }
     });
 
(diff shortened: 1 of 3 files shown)


Naros:

PR #728, opened 2024-08-12
GH-727 Add constant variable support

### LINKED ISSUE #727: Allow for the creation of constants
### Description

In GDScript, when one creates a constant, you define this at the top of the script as:
```python
const SPEED : float = 10.0
```

### Implementation ideas

The same concept can now be done in Orchestrator by either using the "Make ABC" nodes or by defining a variable that you simply explicitly do not use a setter for and assigning its value as the default value.   While both of these work, they're just workarounds.

What probably makes the most sense is to use the `OScriptVariable` implementation and associate a `Constant` attribute to the variable.  

image

When setting a variable as a constant, a setter cannot be spawned for the variable and the variable definition cannot be flagged as an exported value, these are meant to be internal only variables to the script.

In addition, to add clarification to the UI, the variable panel will also reflect the constant flag using the `.C` icon:

image

Naros:
Fixes #727


Naros:

PR #730, opened 2024-08-12
GH-729 Graph grid settings & cache left/right panel layouts

### LINKED ISSUE #729: Add settings for grid pattern, visibility, snapping & cache left/right panel offsets
### Description

Currently, the Grid Pattern, its visibility, and the snapping are not values that persist across restarts.  These should be added to the `Project Settings > Orchestrator > UI > Graph` section so users can control this behavior.

Additionally, the width of the files list and component panels does not persist across editor restarts, so if you've adjusted their widths, this must be done on each Godot execution.  Furthermore, the component panel's width is not synchronized across different loaded scripts, so this can create an odd UX as it is typically expected the right panel's width would be consistent regardless of the file viewed.

### Implementation ideas

_No response_

Naros:
Fixes #729


Naros:

PR #733, opened 2024-08-12
GH-716 Allow moving nodes using arrow keys

### LINKED ISSUE #716: Make Nodes inside EventGraph movable via arrow keys
### Description

Inside the EventGraph it becomes handy to move Nodes via the arrow keys on the keyboard. Currently the arrow keys navigate trough the Godot UI when used.

### Implementation ideas

When the EventGraph is focused selected Nodes should move corresponding to the pressed arrow keys on the keyboard.

Naros:
Fixes #716


Naros:

PR #754, opened 2024-08-16
Support double precision

Naros:
Supersedes #450


Naros:
Hi @fire I have a new idea about double builds, from your point of view do we only need those for linux?

fire:
I don't understand? double precision should be in theory built for all platforms.

Naros:
Ok, that's what I was also thinking, but I wanted to be sure.

To automate the delivery of double-precision builds, I think we can simply build `godot-engine` with the double-precision flag and dump the two API files that are needed for building `godot-cpp` with double-precision.  I tested this process and it works quite well, particularly when we cache the `godot-engine` build.

The main question is how to orchestrate the build on CI.  Obviously, we could set up the CI job to build both precisions in a single matrix and distribute a single plugin archive that has both, but my concern is the archive bundle's size given we'd duplicate each job for Linux, Windows, macOS, Android32, Android64, and Wasm32.

I was thinking since Godot does not officially distribute double builds, would it make sense to distribute a separate plugin artifact that contains specifically just the double-precision builds instead.  The new Update Picker would have to be adjusted to handle this whereas a single distribution wouldn't affect it.

Whats your $0.02 worth @fire ?

fire:
I think there's a gap in how asset library handles double precision so the crucial thing is that I needed a official build of double-precision for my project, but the exact arrangement is up to you.

aaronfranke:
The upstream godot-cpp has renamed this variable to `GODOT_PRECISION`: https://github.com/godotengine/godot-cpp/pull/1583

PR #802, opened 2024-09-08
GH-761 Use new ClassDB methods to fetch property getter/setter

### LINKED ISSUE #761: Support `ClassDB::get_property_setter` and `ClassDB::get_property_getter`
### Description

The current mappings in Orchestrator rely on inspecting the method name and if it starts with `get`, `is`, or `set`, we map it to either be a getter or setter; however, this does not always work because in some cases non-conforming methods are used.  

In Godot 4.4, `ClassDB` has been adjusted to expose `get_property_getter` and `get_property_setter` to obtain these directly from the registered class metadata.

See https://github.com/godotengine/godot/pull/94826

Naros:
Fixes #761 

Without the fix
image

With the fix
image

### DIFF
--- a/src/editor/graph/actions/default_action_registrar.cpp
+++ b/src/editor/graph/actions/default_action_registrar.cpp
@@ -569,8 +569,13 @@ void OrchestratorDefaultGraphActionRegistrar::_register_properties(const String&
             }
         }
 
+        #if GODOT_VERSION >= 0x040400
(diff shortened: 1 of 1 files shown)


Naros:
As this is guarded by `GODOT_VERSION`, this logic will be enabled when we bump to Godot 4.4 as baseline.

PR #813, opened 2024-10-01
Draft - Make dialogue panel more easily extended.

esklarski:
I made a few small changes to make extending the dialogue panel easier.

- Use unique name access in dialogue scene
- Add class_name identifier

This allows adding functionality as easy as:
Screenshot from 2024-10-01 14-26-33

Where I duplicated the default scene, detached the script and added this one and a `TextureRect` for a portrait.

This produces no functional change in the Orchestrator base as far as a user would be concerned.


Naros (approved):
Thanks, @esklarski, this change LGTM.  Congrats on your first contribution 🎉

PR #820, opened 2024-10-10
Goto system / Custom Event node

esklarski:
I've been working on creating a means to begin an orchestration's flow at arbitrary locations, that can be called from GDScript. I've attempted to go about this in two different ways as I've learned the ins and outs of C++ and Orchestrator: via a simple node GotoNode with a name parameter, and via the current function/event system.

For more details, visit the [suggestion thread](https://discord.com/channels/1049457206648655912/1292955976059584563) I created for this topic on the project Discord server.

The first is quite simple, and now that I've figured out how to add inspector properties I might revisit this.
Details: [GotoNode](https://github.com/esklarski/godot-orchestrator/commit/ea1e6728de973b109753bd317d7ad47b9d65ca43)
This hit a block that while I can search through the nodes and find the one with the matching string, I do not know what I can pass to `call()` to start the execution.
Details: [execute_node()](https://github.com/esklarski/godot-orchestrator/commit/f9e99c7f6ca39effeee8e93f7baf1105c7a1f55d)

The second feels like a better solution, but the complexities of the function registration system has me stumped: [CustomEventNode](https://github.com/esklarski/godot-orchestrator/commit/f23dd4207084f6c1eb458f0cab81c1f80c521028)
It seems that it should be able to register a zero parameter function matching it's name on creation, and then deregister that and register a new one when the name is changed. But whenever I try to do so I get crashes and editor hangs 🫤

I'm clearly missing something... and don't understand how editor input works. I need a "user_pressed_enter_in_field" signal, and a way to properly register a function on initialization.


esklarski:
Replaced by #823

PR #829, opened 2024-11-09
Tracking PR for the godot-cpp cmake changes

enetheru:
Hi,

These cmake changes track the changes Ive been making over at [godot-cpp PR:1598](https://github.com/godotengine/godot-cpp/pull/1598)

This is one of the only projects that uses cmake that I can verify my changes with.
If you find them interesting I would appreciate any feedback you can profile.

Cheers,
Samuel.


Naros:
Hi @enetheru did you intend to delete this?

enetheru:
Sorry yeah, I started again.

enetheru:
Here is the second attempt branch https://github.com/enetheru/godot-orchestrator/tree/gdext-4.4
I also made a godot-cpp build profile in this one that cuts the number of godot-cpp sources down from 1K to 350 or so. But I dont know how that will interact with your plugin.

There was one line that was concerning and that was

src/script/serialization/text_loader_instance.cpp:887 which fails to compile

Naros:
I'll take a look. I believe the `text_loader_instance.cpp:887` failure is expected as that line didn't make it into 4.4 :(

Issue #838, opened 2024-12-01
Dialogue: Add Dialogue Message Default Scene as a Project Settings Option


ch0m5:
### Description

Setting the scene to use in every Node is cumbersome. On the documentation, the option to change the Default Scene is proposed, but not yet implemented afaik. This issue aims to resolve this.

![image](https://github.com/user-attachments/assets/b537a6d2-73e6-406f-b36c-c2ebb96ade52)

### Implementation ideas

Add an Orchestrator Project Setting which allows the user to set a .tscn as the Default Scene `Dialogue Message Nodes` use.

PR #840, opened 2024-12-04
Add dialogue message Default Scene to Project Settings

ch0m5:
Adds a setting in Project Settings for the user to set the PackedScene that is instanced by default when a Dialogue Message Node runs in an Orchestrator Script.

Aims to fix issue #838.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@
 
 # Godot Orchestrator
 
-Orchestrator is the ultimate Visual-Scripting solution designed for Godot 4.2+ platform.
+Orchestrator is the ultimate visual scripting solution designed for the Godot 4.2+ platform.  
+
 It revolutionizes Godot game development with its intuitive graph editor interface and hundreds of nodes to build any game logic.
 This add-on empowers you to craft complex behaviors and intricate game logic with unparalleled ease and flexibility, just as you would find in other commercially available game engines.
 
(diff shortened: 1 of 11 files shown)


Naros:
> **Note**: The setting is read every time a Dialogue Message runs (`step()`). This is suboptimal and would be way better to alter the variable holding the path to the default Packed Scene at editor startup and whenever the user changes its value. However, I was unable to find a way to do this. Help or directions to a "Project Settings were altered" method/signal would be much appreciated.

The easiest way would likely be to modify the `OScriptnodeDialogueMessage::instantiate()` method and initialize the value there when the instance is created. This would avoid needing to fetch the value on each call to `step()` to minimize the overhead.

ch0m5:
> > **Note**: The setting is read every time a Dialogue Message runs (`step()`). This is suboptimal and would be way better to alter the variable holding the path to the default Packed Scene at editor startup and whenever the user changes its value. However, I was unable to find a way to do this. Help or directions to a "Project Settings were altered" method/signal would be much appreciated.
> 
> The easiest way would likely be to modify the `OScriptNodeDialogueMessage::instantiate()` method and initialize the value there when the instance is created. This would avoid needing to fetch the value on each call to `step()` to minimize the overhead.
> 
> Just create a `String` variable called `_default_scene` in the instance and then assign that variable in `instantiate`.

The changes have been applied as per your recommendation.

> Btw, could you rebase this PR against `main`?

Done, although it is the first time I rebase a branch (I usually merge). If you see any issue please let me know, otherwise I think the PR should be ready.

Naros:
Hi @ch0m5, so I would expect to only see your commits in the PR and not those of others. The reason I see others is you performed a merge commit rather than a rebase. Could you fix your commit history on this PR?

ch0m5:
> Could you fix your commit history on this PR?

Sorry for the blunder. To be completely honest I am not that proficient with Git (as you might've noticed), so just doing a new PR updated to main and recreate the changes, which are few, would be easier for me rather than "clean up" this faulty rebase.

I created PR #885, which should accomplish the same as this PR but is updated to main's latest changes. If you'd like me to instead fix this PR I can do so, but it'll take me longer due to lack of time and, mainly, experience with Git.

Naros:
Closing in favor of superseded pull request.

Issue #842, opened 2024-12-09
Add Dialogue Audio


ch0m5:
### Description

Add an "Audio" File Pin to `Show Dialogue Message` Nodes, allowing the user to set an Audio for each Dialogue Message in order to seamlessly include audio into the Dialogue Trees built with Orchestrator.

### Implementation ideas

Add an "audio" pin to `Show Dialogue Message` that allows the user to bind an `AudioStream` resource to the Node. When run, the Node passes this path to the `Scene` for the user to handle.

![image](https://github.com/user-attachments/assets/39966aae-fd37-4c24-90b0-f285a89cbf3d)

**Note**: For localization it is assumed the user deals with these issues externally through the Scene, such as using English paths by default and treating it as an ID to instead use a variant path for the appropriate language.

ch0m5:
If this gets merged, I'd like to include a Text-to-Speech feature that plays a generated TTS sfx for Dialogues according to their text when no audio is set.

This would be something you could toggle ON/OFF in Project Settings, and could prove useful in debug/test scenarios where the audio is to be, eventually, replaced by a voice line. Having a placeholder, even of poor quality, helps to visualize the timing of the dialogue and is arguably better than no audio at all.

Naros:
If I recall, there is a hardcoded rate at which text is effectively "typed" in the dialogue, right?

Given that different characters may speak at different speeds due to various characteristics such as dialect or language spoken, I wonder if there needs to be a way to ensure that the audio sound bite, as it plays, is synchronized with the "type" effect?

I haven't worked much in this area to know how other tooling may work or be designed to support this, so just throwing this out there as food for thought.

PR #843, opened 2024-12-09
Add audio for Dialogue Message nodes

ch0m5:
Aims to solve #842.


src/script/nodes/dialogue/dialogue_message.cpp
```diff
@@ -57,7 +57,7 @@ class OScriptNodeDialogueMessageInstance : public OScriptNodeInstance
             return selection == -1 ? 0 : selection;
         }
 
-        Variant scene_file = p_context.get_input(2);
+        Variant scene_file = p_context.get_input(3);
```
Naros, line 60: Unfortunately, this change won't be backward compatible. You can't change the pin order or indices without there being issues. In short, this means that any older orchestrations won't connect their pins properly.

You may need to take a look at `_upgrade` or `reallocate_pins_during_reconstruction` to deal with correctly handling the upgrade for existing dialogue message nodes in older orchestrations.

Naros (changes requested):

PR #853, opened 2025-01-17
Node Metadata Access to DialogueMessage and DialogueChoice Nodes 

### LINKED ISSUE #852: Add Node Metadata Access to DialogueMessage and DialogueChoice Nodes
### Description

Currently, the Dialogue-related Nodes (Message and Choice) have a very pre-defined and limited amount of data, which creates hurdles when you want a dialogue system that goes beyond basic functions. For example:
- The screen should shake when a Choice is hovered.
- A Choice must be visible, but not clickable.
- A Choice leads to a special event, like combat. We want to add the "combat" metadata to edit the Choice's text at runtime.

Basically, there needs to be a way to send custom data to the Dialogue Scene that we can interpret in the GDScript game environment and act accordingly, and given that Orchestrator Nodes can hold _metadata_, which is saved on the .torch file too, it would be an easy way to add this layer of customization.

### Implementation ideas

The implementation would consist in adding aditional `"metadata"` keys to the `dialogue_data` Dictionary within the Dialogue Scene.

For the DialogueMessage, a `"metadata"` key with a Dictionary value containing all metadata <String, Variant> pairs.

For the DialogueChoice metadata, it would require a bit of a refactor. `"options"` should no longer be an Array[String], but rather an Array[Dictionary], where each value is a data-package of each choice:
- Text
- Visibility
- Metadata Dictionary

P.D.: This method would also improve how "Option Visibility" is handled. Currently, a non-visible Option is not passed to the `"options"` Array, which is not ideal since it is lost data. All options, visible and otherwise, should be passed to the Godot Scene and let the user handle them as they wish.

ch0m5:
Addresses #852.

Note that this PR changes the `"options"` Array passed to the Godot Scene from essentially `Array[String]` to `Array[Dictionary]`, where each Dictionary is a data-package for each Choice, rather than just its `text` (the prior case).

Therefore, it _changes_ how Choice data is sent and interpreted from the GDExtension to GDScript, therefore affecting compatibility with any Orchestrations with Dialogue created prior to this PR.

### DIFF
--- a/src/script/nodes/dialogue/dialogue_message.cpp
+++ b/src/script/nodes/dialogue/dialogue_message.cpp
@@ -45,6 +45,19 @@ class OScriptNodeDialogueMessageInstance : public OScriptNodeInstance
         return FileAccess::file_exists(vformat("%s.remap", p_path));
     }
 
+    Dictionary _get_object_metadata(Object obj) const
+    {
+        TypedArray<String> keys = obj.get_meta_list();
+        Dictionary metadata;
+        for (int i = 0; i < keys.size(); i++)
+        {
+            StringName key = keys[i];
+            Variant value = obj.get_meta(key);
+            metadata[key] = value;
+        }
+        return metadata;
+    }
+
 public:
     int get_working_memory_size() const override { return 1; }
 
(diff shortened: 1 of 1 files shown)


Naros:
Hi @ch0m5, could you rebase rather than doing a merge commit

PR #885, opened 2025-02-12
Added default_scene to ProjectSettings (The Sequel)

ch0m5:
Adds a setting in Project Settings for the user to set the PackedScene that is instanced by default when a Dialogue Message Node runs in an Orchestrator Script.

Aims to fix issue #838.

**Note**: This is a re-creation of the Branch in PR #840. A rebase was executed incorrectly and the commit history got all messed up. This PR aims to accomplish the exact same goal with none of the clutter created by the aforementioned blunder.


Naros:
Hi @ch0m5 I hope this doesn't create any issues for your team but I am going to modify the default scene as follows:

image

I'd prefer we put this under its own section.

PR #916, opened 2025-03-09
Fix compiler warnings

Naros:


Naros:
Close, already applied directly.

PR #931, opened 2025-03-23
Introduce dual builds - editor/template_release

Naros:
Introduce dual builds based on `editor` and `template_release` targets for `godot-cpp`.   This will provide the ability for editor specific behavior to be used inside the editor runtime, and to provide slimmer builds for exported games.


Naros:
Introduced in separate PRs.

Issue #911, opened 2025-03-03
Ability to duplicate functions


Trico-Everfire:
### Description

It would be nice if you could duplicate functions. Let's say you have two movement systems that use similar (but not identical) logic you may wanna copy most of the logic to the other function. So it'd be nice if you could duplicate a function.

Naros:
Are you looking to duplicate the function & its nodes or just the function stub, i.e. the name, arguments, and return?

Trico-Everfire:
Function with it's nodes.

PR #962, opened 2025-07-01
GH-956 add functionality to duplicate functions and variables in graphs

### LINKED ISSUE #956: Add a context menu entry to to duplicate a variable.
### Description

Right now if you need multiple variables of the same type, you need to create them one by one. In a text editor one would duplicate the line and just change what needs changing. So it would be nice if there was duplicate context menu entry on variables in the context menu:

Image

This would just create a duplicate of the variable with all settings copied and just pick a unique name for it (e.g. by appending a number). 

### Implementation ideas

_No response_

derkork:
Fixes #956 and #911

I had a stab at this as it seemed reasonably easy to implement and gave me a chance to get acquainted with the code base. I have tried to follow the code style and idioms used in the code base, but I might have added stuff in places where it's maybe not supposed to go, so please let me know if any changes are required to get this merged.

### DIFF
--- a/src/editor/component_panels/variables_panel.cpp
+++ b/src/editor/component_panels/variables_panel.cpp
@@ -118,6 +118,7 @@ String OrchestratorScriptVariablesComponentPanel::_get_remove_confirm_text(TreeI
 
 bool OrchestratorScriptVariablesComponentPanel::_populate_context_menu(TreeItem* p_item)
 {
+    _context_menu->add_icon_item(SceneUtils::get_editor_icon("Duplicate"),"Duplicate", CM_DUPLICATE_VARIABLE);
     _context_menu->add_icon_item(SceneUtils::get_editor_icon("Rename"), "Rename", CM_RENAME_VARIABLE, KEY_F2);
     _context_menu->add_icon_item(SceneUtils::get_editor_icon("Remove"), "Remove", CM_REMOVE_VARIABLE, KEY_DELETE);
     return true;
@@ -133,7 +134,16 @@ void OrchestratorScriptVariablesComponentPanel::_handle_context_menu(int p_id)
         case CM_REMOVE_VARIABLE:
             _confirm_removal(_tree->get_selected());
             break;
+        case CM_DUPLICATE_VARIABLE:
+            _duplicate_variable(_tree->get_selected());
     }
+}
+
+void OrchestratorScriptVariablesComponentPanel::_duplicate_variable(TreeItem* p_item)
+{
+    Ref<OScriptVariable> duplicate = _orchestration->duplicate_variable(_get_tree_item_name(p_item));
+    update();
+    _find_child_and_activate(duplicate->get_variable_name());
 }
 
 bool OrchestratorScriptVariablesComponentPanel::_handle_add_new_item(const String& p_name)
--- a/src/editor/component_panels/variables_panel.cpp
+++ b/src/editor/component_panels/variables_panel.cpp
@@ -142,8 +142,11 @@ void OrchestratorScriptVariablesComponentPanel::_handle_context_menu(int p_id)
 void OrchestratorScriptVariablesComponentPanel::_duplicate_variable(TreeItem* p_item)
 {
     Ref<OScriptVariable> duplicate = _orchestration->duplicate_variable(_get_tree_item_name(p_item));
-    update();
-    _find_child_and_activate(duplicate->get_variable_name());
+    if (duplicate.is_valid())
+    {
+        update();
+        _find_child_and_activate(duplicate->get_variable_name());
+    }
 }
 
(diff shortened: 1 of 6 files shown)


src/orchestration/orchestration.cpp
```diff
@@ -616,13 +620,95 @@ Ref<OScriptFunction> Orchestration::create_function(const MethodInfo& p_method,
...
     return function;
 }
+Ref<OScriptFunction> Orchestration::duplicate_function(const StringName& p_name, bool p_include_code)
+{
+    ERR_FAIL_COND_V_MSG(_has_instances(), nullptr, "Cannot duplicate functions, instances exist.");
+    ERR_FAIL_COND_V_MSG(!has_function(p_name), nullptr, "No function exists with the name: " + p_name);
+
+    Ref<OScriptGraph> old_graph = find_graph(p_name);
+    Ref<OScriptFunction> old_function = find_function(p_name);
+    Ref<OScriptFunction> result;
```
Naros, line 638: I might suggest using `new_function` here as the name, to just avoid confusion with the result node

src/orchestration/orchestration.cpp
```diff
@@ -616,13 +620,95 @@ Ref<OScriptFunction> Orchestration::create_function(const MethodInfo& p_method,
...
+
+    // make a unique name for the new function
+    String new_name = NameUtils::create_unique_name(p_name, get_function_names());
+
+    // make a graph
+    Ref<OScriptGraph> graph = create_graph(new_name, OScriptGraph::GF_FUNCTION | OScriptGraph::GF_DEFAULT);
+
+
+    // duplicate each node, make a dictionary that maps old node IDs to new node IDs
+    Dictionary node_id_map;
```
Naros, line 648: I would suggest using `HashMap<int, int>` rather than a `Dictionary`. It will bypass the coercion into a Variant and will perform better for larger graphs that get copied.

src/orchestration/orchestration.cpp
```diff
@@ -616,13 +620,95 @@ Ref<OScriptFunction> Orchestration::create_function(const MethodInfo& p_method,
...
+            MethodInfo mi = old_function->get_method_info();
+            mi.name = new_name;
+            context.method = mi;
+            entry->initialize(context);
+            result = entry->get_function();
+        }
+    }
+
+    // now restore connections
+    for (const OScriptConnection& old_connection : old_graph->get_connections())
```
Naros, line 694: If we are not duplicating the function code, we really can skip this step, and in such cases, I would just manually wire the function entry and result nodes together directly.

src/orchestration/orchestration.cpp
```diff
@@ -774,6 +860,28 @@ Ref<OScriptVariable> Orchestration::create_variable(const StringName& p_name, Va
...
+
+    Ref<OScriptVariable> result = create_variable(new_name, original->get_variable_type());
+    // copy all properties
+    result->set_category(original->get_category());
+    result->set_default_value(original->get_default_value());
+    result->set_classification(original->get_classification());
+    result->set_custom_value_list(original->get_custom_value_list());
+    result->set_constant(original->is_constant());
+    result->set_description(original->get_description());
+    result->set_exported(original->is_exported());
```
Naros, line 880: If there isn't a method on `OScriptVariable` yet, I would suggest adding a `copy_persistent_state` method where you pass in the old variable and it copies the state into the new. There are a couple places I think this can be viable in my rework, so having this in one place makes good sense for future property additions/removals.

I would also guard in case `result` is not valid so we don't throw a segfault.

src/editor/component_panels/variables_panel.cpp
```diff
@@ -133,9 +134,18 @@ void OrchestratorScriptVariablesComponentPanel::_handle_context_menu(int p_id)
...
+        case CM_DUPLICATE_VARIABLE:
+            _duplicate_variable(_tree->get_selected());
     }
 }
 
+void OrchestratorScriptVariablesComponentPanel::_duplicate_variable(TreeItem* p_item)
+{
+    Ref<OScriptVariable> duplicate = _orchestration->duplicate_variable(_get_tree_item_name(p_item));
+    update();
+    _find_child_and_activate(duplicate->get_variable_name());
```
Naros, line 146: I would guard these two lines with a `if (duplicate.is_valid())` check for good measure.

Naros (commented):
Hi @derkork, great first pass at this - perfect. I've left a few comments.

src/orchestration/orchestration.cpp
derkork: I used this function, but it also copies over the name, which I'm not sure if this is ok. I added a line to restore the name, but it feels somewhat hacky.

derkork:
I have implemented the requested changes .

PR #1003, opened 2025-07-12
GH-1002 Separate data model from script extension

### LINKED ISSUE #1002: Create separation from script and data model
### Description

Currently, the `OScript` class extends the `Orchestration` class, which means that both the script and the model are tightly coupled. This worked well when there was only the script module; however, as we move to introduce macro libraries, these two need a clear separation.

### Implementation ideas

I propose moving from an inheritance-based approach where `OScript` uses the `Orchestration` class as a parent type, to which we move `Orchestration` to a `RefCounted` object, and let `OScript` hold a reference. This shift provides a myriad of benefits:

* We can share `Orchestration` across `OScript` and `OMacroLibrary`. 
* We can treat `Orchestration` like the "source code".
* Allows for easier reloading of a script that has always been problematic with inheritance.

In addition, this means that `Orchestration` and its associated graphs, nodes, functions, and variable classes will now represent what is often referred to as the _data model_. These classes should all be relocated into the `orchestration/` directory to show again a clear distinction between objects in `script/` and `orchestration/`.

Naros:
Fixes #1002

### DIFF
--- a/src/script/serialization/serialization_binary.cpp
+++ b/src/script/serialization/serialization_binary.cpp
@@ -24,6 +24,7 @@
 #include <godot_cpp/classes/dir_access.hpp>
 #include <godot_cpp/classes/file_access.hpp>
 #include <godot_cpp/classes/project_settings.hpp>
+#include <godot_cpp/classes/resource_loader.hpp>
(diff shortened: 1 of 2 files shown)


Naros:
Fixed in #1141

PR #1088, opened 2025-11-29
Improve CI Web Build Cache

Naros:
This will run sccache in local server mode, and allow the actions/cache step to handle the build cache for web builds.


Naros:
This is more complicated than it's worth, leaving as-is for now.

PR #1145, opened 2026-01-01
GH-1134 Do not autowire when prior drag canceled

### LINKED ISSUE #1134: Adding new action after dragging from a pin, then cancelling, still tries connecting from the pin
### Describe the bug

If you drag from a pin and drop on an empty spot to pull up the new graph action dialog, but then cancel the action either by pressing esc or clicking the cancel button, and then right click to create a new action separately, when the new node is added it incorrectly gets connected to the pin that you dragged earlier.

### Expected behavior

Creating new node from right click shouldn't pick up any connections from pins.

### Actual behavior

If added after dragging out a pin, cancelling, and then separately adding a new node the pin is still connected as if you had just dragged it.

### How to Reproduce?

see description

### Godot full version

4.5.1

### Orchestrator version

2.3.2

### Additional information

_No response_

Naros:
Fixes #1134


Naros:
Backported to only 2.3 and 2.2.

PR #1152, opened 2026-01-02
GH-1151 Add ui/nodes/connection_hotzone_scale option

### LINKED ISSUE #1151: Add option to scale the graph node pin hot zone area
### Description

The current node pin hot zone area is based on the editor's theme for `port_hotzone_outer_extent` and `port_hotzone_inner_extent`, but changing these globally would impact other plugins and widgets and so we should likely implement a custom option for this.

I propose introducing a new **UI > Nodes > Connection Hotzone Scale** option to select 100%, 150%, and 200%. During the connection wire drag operation to/from a node port, this will increase the `Rect2` area around the port by 1.5 to 2 times of the default theme value, making it easier to connect wires to ports when close to the pin.

### Implementation ideas

_No response_

Naros:
Fixes #1151


Naros:
Applied to 2.3, 2.2, and 2.1 only.  There is an API incompatibility with 2.0.

PR #1157, opened 2026-01-04
GH-1156 Fix autowire eligibility algorithm

### LINKED ISSUE #1156: When auto-wire dialog choices are shown, class type is not considered
### Describe the bug

When dragging from a pin that represents object A and selecting a menu that takes object B plus as its context and an `Any` / `Variant` argument, the autowire eligibility window will show not only the variant as eligible but also object B where object A and B are incompatible.

### Expected behavior

Autowire dialog should only match against
1. Variant arguments
2. If the dragged pin is a class, class types that are the same or parent types of the context pin.
3. If the dragged pin is an enum/bitfield, arguments that match the enum/bitfield or integer-arguments.
4. Same basic types, i.e. PackedColorArray maps to PackedColorArray but not PackedStringArray.

### Actual behavior

The autowire dialog includes non-eligible pins for connection.

### How to Reproduce?

1. Drop an Input singleton 
2. Drag from the output pin and select `Track Set Key Value` from `Animation`.
3. Autowire dialog presents two choices (`Animation` context pin and the `Variant` value pin).

### Godot full version

_No response_

### Orchestrator version

_No response_

### Additional information

_No response_

Naros:
Fixes #1156

### DIFF
--- a/src/editor/graph/graph_node_pin.cpp
+++ b/src/editor/graph/graph_node_pin.cpp
@@ -193,7 +193,11 @@ bool is_numeric(Variant::Type p_type)
 
 bool OrchestratorGraphNodePin::can_accept(OrchestratorGraphNodePin* p_pin)
 {
-    return _pin->can_accept(p_pin->_pin);
+    ERR_FAIL_NULL_V(p_pin, false);
+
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.3 and 2.2. Not eligible for 2.1 and 2.0.

PR #1170, opened 2026-01-18
GH-1168 Make mouse visible when dialogues are shown

### LINKED ISSUE #1168: Dialogue Message scene does not show mouse when hidden
### Describe the bug

The `OScriptNodeDialogueMessage` script node should automatically show the mouse if its captured.

### Expected behavior

The mouse should always be visible when the dialogue message scene is shown.

### Actual behavior

The mouse state is left as-is.

### How to Reproduce?

1. Create a game that hides the mouse.
2. Have a trigger that shows the dialogue scene
3. Mouse is never made visible

### Godot full version

n/a

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1168 

The current design of the dialogue message script is that it does not take into account if the game's mouse has been made hidden. This change is designed to cache the current mouse state, and if it's hidden, make it visible temporarily, and return it as hidden after the dialogue window is closed.

### DIFF
--- a/project/addons/orchestrator/scenes/dialogue_message.gd
+++ b/project/addons/orchestrator/scenes/dialogue_message.gd
@@ -39,6 +39,7 @@ var selection : int
 
 var _current_tween : Tween
 var _current_choices : Dictionary
+var _mouse_mode : Input.MouseMode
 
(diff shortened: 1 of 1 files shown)


Naros:
Backported to 2.3, 2.2, 2.1, and 2.0

Issue #1141, opened 2025-12-31
Make scripting runtime engine bytecode-based


Naros:
### Description

The virtual machine used by Orchestrator up to this point has worked well, but it has left the plugin in a position where key components such as the script, data model, and runtime are too tightly coupled. This coupling makes it challenging to introduce specific features or changes. To address this problem and several performance issues we've identified, the proposal is to rewrite the scripting virtual machine to use a fully bytecode-based solution, similar to GDScript.

This process will require multiple changes to achieve the goal.

**Decouple `Orchestration` from `OScript`**

One of the challenges with the user interface is that `Orchestration` is an interface implemented by `OScript`. This means the script implementation and the data model are identical. This also introduces challenges with reloading content and performing actions independent of the `Script` lifecycle.

Moving toward composition rather than inheritance enables reusing `Orchestration` in macro libraries.

**Compile-Once Strategy**

This is where the bulk of the changes come in. Up until now, when an `OScriptInstance` (the runtime object for a script) is created, a series of passes occur that generate dozens to hundreds of subobjects to represent the lifecycle of that script for a single node in your scene.  If a script is used by more than one node, it is evaluated N times (where N is the number of nodes that use it).

This model doesn't scale well for large scenes or games that want to harness the power of composition and reuse. 

Orchestrator will shift to a model that aligns with GDScript, where the script is evaluated when its source is loaded off disk or as a reload is triggered. The script's executable details are managed statelessly, enabling a compile-once strategy and reusing the compiled objects across all nodes that use the script in the scene. This would make the `OScriptInstance` (the runtime instance object for a script per node) extremely lightweight, storing only variable state.  This too mimics how GDScript works.

**Multi-pass Compilation**

The current compilation uses a single pass to transform the `Orchestration` and its nodes to the runtime representation. This results in a very tight coupling between the data model and the runtime. Moving forward, we'd like to shift to a paradigm where the runtime is independent of the data model, and to that end, we'll leverage a multi-pass compilation process.

First, the `Orchestration` and its sub-resources will be transformed into an Abstract Syntax Tree (AST). This conversion provides several key advantages, most notably enabling us to migrate from the data model to an intermediate, entirely agnostic model. In addition, ASTs are extremely good at representing code, particularly in text-based languages. 

We'll use the AST model to perform several passes over the tree. These passes use standard techniques to evaluate the tree, identify dead code, and perform expression reduction. The goal of these passes is to reduce code footprint while improving script performance at runtime.

[...]

PR #1177, opened 2026-01-19
GH-1176 Fix script error message instance resolution

### LINKED ISSUE #1176: Script error message texts related to object-based operations always reports a 'null instance' in the error text
### Describe the bug

When a script that calls a method or operates on an object instance, if the call fails and the script virtual machine throws an error, it reports that the error occurred on a 'null instance' rather than giving the object instance's name.

### Expected behavior

Object instance names should be provided

Image

### Actual behavior

Always reports 'null instance.

Image

'

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

A regression related to GH-1141.

Naros:
Fixes #1176, a regression of GH-1141.

### DIFF
--- a/src/script/script_vm.cpp
+++ b/src/script/script_vm.cpp
@@ -65,7 +65,7 @@ static String _get_var_type(const Variant* p_var) {
 
     if (p_var->get_type() == Variant::OBJECT) {
         bool was_freed;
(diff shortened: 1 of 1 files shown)


Naros:
Backported to 2.3, and 2.2

PR #1179, opened 2026-01-19
GH-1173 Fix `GDE::Variant::evaluate` to return result

### LINKED ISSUE #1173: Exported string values are constantly cleared
### Describe the bug

For script types that export a string-based property, when typing the content, the field will be cleared after every few characters.

### Expected behavior

The field should not be cleared.

### Actual behavior

The field is cleared periodically.

### How to Reproduce?

1. Create a script and export a string-based variable. 
2. Attach script to a node.
3. Click the node
4. Type in the exported property field in the inspector

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

This is a regression related to GH-1141. The logic inside `GDE::Variant::evaluate` does not return the result of the operator evaluation call, but rather whether the call was valid only.

Naros:
Fixes #1173, a regression to GH-1141

### DIFF
--- a/src/core/godot/variant/variant.cpp
+++ b/src/core/godot/variant/variant.cpp
@@ -149,10 +149,9 @@ Variant GDE::Variant::evaluate(godot::Variant::Operator p_operator, const godot:
     return result;
 }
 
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.3 and 2.2

PR #1181, opened 2026-01-19
GH-1171 Correctly enqueue print string text overlay

### LINKED ISSUE #1171: Print String Overlay overrides previous text
### Describe the bug

Before GH-1141, the print string UI overlay would append additional text below existing text using a `VBoxContainer`; however, since the update the text is always appended over the other text, making it difficult to read.

### Expected behavior

When not providing any keys, text should be appended to new rows in the `VBoxContainer`.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1171, regression from GH-1141.


Naros:
Backported to 2.3 and 2.2

Issue #967, opened 2025-07-04
Support Godot 4.4


Naros:
### Description

Add Godot 4.4. ABI compatibility

PR #1183, opened 2026-01-19
GH-1182 Correctly bind struct pin `text_submitted` signal

### LINKED ISSUE #1182: Struct-based pin throws error modifying default pin values
### Describe the bug

When modifying a struct-based pin (i.e. Vector2, Vector3, Vector4, etc), and modifying the default values rather than connecting an input node, the change of the value throws an exception:

```
ERROR: Error calling from signal 'text_submitted' to callable: '': Method expected 0 argument(s), but called with 1.
   at: emit_signalp (core/object/object.cpp:1378)
```

### Expected behavior

No error should be thrown.

### Actual behavior

Error thrown.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

This is a regression from GH-967, the UI rework.

Naros:
Fixes #1182, regression to GH-967


Naros:
Backported to 2.3 and 2.2.

PR #1185, opened 2026-01-22
GH-1172 Fix OScriptParser handling variadic function args

### LINKED ISSUE #1172: Variadic arguments on pure function calls are ignored
### Describe the bug

As an example, the `str()` function is designed to support variadic arguments; however, the `build_pure_call` expression handler does not take into account the fact that the visual script node may have more input pins than the method signature due to the variadic nature of the method signature. This leads to arguments being ignored in the method call.

### Expected behavior

All variadic arguments should be consumed.

### Actual behavior

Only the first argument is consumed.

### How to Reproduce?

Create a `str` node that takes two input arguments and print the contents. The function call will only output the first argument and ignore any of the remaining arguments.

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

This is a regression introduced with the new bytecode virtual machine in GH-1141.

Naros:
Fixes #1172

### DIFF
--- a/src/script/parser/parser.cpp
+++ b/src/script/parser/parser.cpp
@@ -531,6 +531,29 @@ OScriptParser::BinaryOpNode* OScriptParser::create_binary_op(VariantOperators::C
     return binary_op_node;
 }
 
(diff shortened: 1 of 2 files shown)


Naros:
Applied to 2.3 and 2.2

PR #1191, opened 2026-01-24
GH-1190 Rebind parser bindings after clear

### LINKED ISSUE #1190: OScriptParser handler bindings can lead to crashes
### Describe the bug

In the `OScriptParser` class, there is a series of expression and statement handlers that are registered, each of which stores a reference to the `this` pointer through a lambda call. The issue is that inside `OScriptParser::clear`, we use a hack to reset the parser's state in-place; however, this does not rebind the handlers, leading to a dereference of a no longer valid `this` pointer in the lambda bindings.

### Expected behavior

There should be no crash.

### Actual behavior

OScriptParser potentially can crash

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

It would seem that as long as the bindings are rebound when calling clear, that should fix the issue.

Naros:
Fixes #1190

### DIFF
--- a/src/script/parser/parser.cpp
+++ b/src/script/parser/parser.cpp
@@ -163,6 +163,84 @@ static String _get_annotation_error_string(const StringName& p_annotation_name, 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 /// OScriptParser
 
+void OScriptParser::bind_handlers() {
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.3 and 2.2

PR #1193, opened 2026-01-24
GH-1192 Fix preload to work with script global classes

### LINKED ISSUE #1192: Preload node does not correctly handle script classes
### Describe the bug

When defining a GDScript class that extends Resource, the `Preload` node does not correctly resolve the class details and allows accessing methods/properties on the class instance.

### Expected behavior

Should be able to access methods/properties/signals on resource instances.

### Actual behavior

Treats return only as a `Resource` native class.

### How to Reproduce?

1. Create a GDSscript class that extends Resource:
```
class_name PlayerData
extends Resource

@export var score : int = 0

func get_score() -> int:
  return score
```

2. Create an instance of the `PlayerData` resource as `player.tres`.
3. Drag `player.tres` onto the graph canvas and drag away from the node; no score-related attributes are available.
4. 

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2/2.1/2.0

### Additional information

_No response_

Naros:
Fixes #1192

### DIFF
--- a/src/common/property_utils.cpp
+++ b/src/common/property_utils.cpp
@@ -16,8 +16,12 @@
 //
 #include "common/property_utils.h"
 
-#include "godot_cpp/templates/hash_map.hpp"
-#include "string_utils.h"
+#include "common/string_utils.h"
+#include "core/godot/scene_string_names.h"
+
+#include <godot_cpp/classes/resource_loader.hpp>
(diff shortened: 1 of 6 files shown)


Naros:
Backported to 2.3 and 2.2

PR #1199, opened 2026-01-24
GH-1198 Fix New Object to work with script classes

### LINKED ISSUE #1198: Using New Object node does not work with script classes
### Describe the bug

When defining a class, the "New Object" node refers to the script path and not the class name. This leads to the inability to create a new instance of the class.

### Expected behavior

The "New Object" node should refer to the class name.

### Actual behavior

_No response_

### How to Reproduce?

1. Create a Godot GDScript as follows

```
class_name PlayerData
extends Resource

@export var score : int = 0

func get_score() -> int:
  return score
```

Then use the "New Object" node to create an instance of `PlayerData`.  While the class is picked properly in the dialog, the node refers to the class by its script path name and not the class.

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1198

### DIFF
--- a/src/script/nodes/memory/memory.cpp
+++ b/src/script/nodes/memory/memory.cpp
@@ -18,6 +18,7 @@
 
 #include "common/property_utils.h"
 #include "common/version.h"
+#include "script/script_server.h"
 
 #include <godot_cpp/classes/engine.hpp>
 #include <godot_cpp/classes/node.hpp>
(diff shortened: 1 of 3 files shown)


Naros:
Backported to 2.3 and 2.2.

PR #1200, opened 2026-01-24
GH-261 Support orchestration script classes

### LINKED ISSUE #261: Support for creating classes

Naros:
Fixes #261

### DIFF
--- a/src/editor/graph/graph_panel.cpp
+++ b/src/editor/graph/graph_panel.cpp
@@ -48,6 +48,7 @@
 #include "script/nodes/data/dictionary.h"
 #include "script/nodes/editable_pin_node.h"
 #include "script/nodes/functions/call_member_function.h"
+#include "script/nodes/functions/call_parent_function.h"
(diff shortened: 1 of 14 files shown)


Naros:
Backported to 2.3 and 2.2

PR #1203, opened 2026-01-24
GH-1202 Fix parser error when functions return variants

### LINKED ISSUE #1202: Parser error - Functions that return variants (any) fail to be parsed
### Describe the bug

When a function returns an _any_ (Variant) value, the parser will throw an error.
```
Node 123456 - Return Node / Function Result: A void function cannot return a value.
```

### Expected behavior

Parser should not throw an error.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1202


Naros:
Backport to 2.3 and 2.2.

PR #1204, opened 2026-01-24
GH-932 Fix overriding parent functions

### LINKED ISSUE #932: Can't seem to override methods.
### Describe the bug

I created a base class in GDscript (message_base.gd) Defined a function send_message(string: message). Then created a new orchestration, extended message_base but my function is not available in the node list. Sorry if this is just me being an idiot. 

### Expected behavior

Should be able to add a node for send_message (similar to _process etc.) 

### Actual behavior

Doesn't show up in the list... 

### How to Reproduce?

Create script 

`class_name message_base
extends Node
func send_message(data: string)
       pass
`
Create new orchestration
Extend message_base
Right click 
Search for send_message

### Godot full version

4.3

### Orchestrator version

2.1

### Additional information

_No response_

Naros:
Fixes #932

### DIFF
--- a/src/editor/graph/graph_panel.cpp
+++ b/src/editor/graph/graph_panel.cpp
@@ -1159,6 +1159,42 @@ bool OrchestratorEditorGraphPanel::_create_new_function(const String& p_name, bo
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


Naros:
Bacakported to 2.3 and 2.2.

PR #1207, opened 2026-01-24
GH-1201 Always place event nodes in event graphs

### LINKED ISSUE #1201: Event nodes can be mistakenly added to function graphs
### Describe the bug

If a function graph is open, an event override is added to the function graph and not the open event graph.

### Expected behavior

Event nodes should always be added to the event graph. The logic should always prefer the current open event graph if an event graph is focused; otherwise falling back to the default EventGraph.

### Actual behavior

Event nodes are added to function graphs if focused.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

A regression related to GH-967.

Naros:
Fixes #1201


Naros:
Backported to 2.3 and 2.2.

PR #1214, opened 2026-01-25
GH-1213 Fix freezes with print string node in tool scripts

### LINKED ISSUE #1213: Tool script freezes Godot
### Describe the bug

When creating a new script, it may be natural to drop a "Print String" node into a `_ready` or similar function to get past the "cannot have an empty function body" error in a tool-based script. Doing this will cause the editor to freeze when "Print to Screen" is enabled.

**Workarounds**
1. Leave "Print to Screen" unchecked
2. Use the `prints` or similar built-in functions and avoid "Print String" 

### Expected behavior

The editor should not freeze

### Actual behavior

Editor freezes every 2 seconds.

### How to Reproduce?

1. Create an orchestration
2. Set it as tool
3. Add print string node to one of the functions

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

To address this, the UI manipulation should likely be omitted when the node runs in a tool-script.

Naros:
Fixes #1213


Naros:
Backported to 2.3 and 2.2

PR #1217, opened 2026-01-25
GH-1216 Fix populating script list on startup

### LINKED ISSUE #1216: Script list fails to load when no prior project state exists
### Description

Orchestrator is designed to restore layout and other plugin-specific values when the editor starts. However, if the project is empty or no specific layout data for the plugin has been saved, the script list will not be populated.

The workaround solution is to open a second script and restart the editor, making sure that you save the project first. This will force there to be state to be restored, so that the script list is populated correctly.

### Implementation ideas

_No response_

Naros:
Fixes #1216


Naros:
Backported to 2.3 and 2.2

PR #1218, opened 2026-01-25
GH-1215 Add soft reload tool script file menu option

### LINKED ISSUE #1215: Add tool script reload menu option like Script tab
### Description

In the Script tab, there is an option to force the soft reload of tool scripts. We should introduce a similar feature for the orchestration tool scripts.

### Implementation ideas

_No response_

Naros:
Fixes #1215


Naros:
Backported to 2.3 and 2.2

PR #1220, opened 2026-01-25
GH-1219 Add tool script hint and background color tint

### LINKED ISSUE #1219: Add tool script background color overlay and hint
### Description

For new users who may be experimenting with tool scripts, they may be unaware that mistakenly pressing `Ctrl+S`, the standard keybind for saving a scene will save any tool script modications to scenes. This is standard Godot behavior, but to help avoid confusion, tool scripts should have 

- A unique graph background
- A hint at the bottom of the screen

### Implementation ideas

Add something like this

Image

And can be modified like

Image

Naros:
Fixes #1219

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -24,6 +24,7 @@
 
 // Helper setting layouts
 #define BOOL_SETTING(n,v) PropertyInfo(Variant::BOOL, n), v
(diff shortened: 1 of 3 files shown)


Naros:
Backported to 2.3 and 2.2.

PR #1227, opened 2026-01-26
GH-1226 Fix parser to set variable export info correctly

### LINKED ISSUE #1226: Exporting variables of object types lead to parser error
### Describe the bug

When exporting object type variables, the parser reports the following error:
```
ERROR: Parser Error: Export type can only be built-in, a resource, a node, or an enum.
   at: OScript::reload (res://tests/tool/rotate.torch:-1)
```

### Expected behavior

Specific types, such as `CompressedTexture2D`, which extends `Resource`, should be exportable without errors

### Actual behavior

When exporting such types, the parser fails with an error.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1226

A regression of GH-1141


Naros:
Backported to 2.3 and 2.2

PR #1234, opened 2026-01-31
GH-1062 Add Godot class-specific script templates

### LINKED ISSUE #1062: Feature request - character body 3d/2d template code
When creating a new orchestrator script, I noticed that there are templates available just just GD script, but when creating the script, the script is blank. There is no example code.

I'm requesting to have a feature for beginners to simply create a template script with example code so they can see what the code looks like when it's working, and this should allow beginners to build off of the template.

Naros:
Fixes #1062


Naros:
Backported to 2.3 and 2.2

PR #1251, opened 2026-02-06
GH-1247 Correctly register static functions

### LINKED ISSUE #1247: Some `Variant` static methods require an unnecessary input
### Describe the bug

`Variant` static methods like `Quaternion.from_euler()` and `Color.from_hsv()` require an input value when placed as graph nodes, in contrast to GDScript where they do not. They also only appear in the list when dragging a pin of their respective types.

### Expected behavior

The static methods appear without pin-dragging and do not have a required input, as is the case with `Tween.interpolate_value()`.

### Actual behavior

The static methods only appear when pin-dragging and cause invalidation unless an existing `Variant` pin is connected.

### How to Reproduce?

_No response_

### Godot full version

v4.6.stable.official [89cea1439]

### Orchestrator version

v2.4.rc1.official [1cd4573f9]

### Additional information

_No response_

Naros:
Fixes #1247


Naros:
Backported to 2.3 and 2.2.

PR #1254, opened 2026-02-08
GH-1246 Rework Break for extra Variant properties

### LINKED ISSUE #1246: `Color` properties are not exposed
### Describe the bug

`Color` properties have no setters in the All Actions menu. As a result, you cannot set properties like `color.h`.

Image

### Expected behavior

The list of available graph nodes should include setters for individual `Color` properties

Or:
Additional "Break" nodes should be available to use with the existing static constructors like `Color.from_hsv()`

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

v4.6.stable.official [89cea1439]

### Orchestrator version

v2.4.rc1.official [1cd4573f9]

### Additional information

_No response_

Naros:
Fixes #1246 

The break nodes were previously designed to only render specific Variant properties on Color, Plane, and AABB that matched the arguments in the variant type's constructor. This meant that for these types, only limited properties could be accessed.

In addition, the break node for AABB exposes end and for Plane the break node exposes normal.

image

The rework involves creating 4 specific break nodes for Color: RGBA, RGBA8, HSV, and OK_HSL

image

### DIFF
--- a/src/editor/actions/introspector.cpp
+++ b/src/editor/actions/introspector.cpp
@@ -854,12 +854,43 @@ Vector<Ref<OrchestratorEditorIntrospector::Action>> OrchestratorEditorIntrospect
                     .build());
             }
 
-            actions.push_back(
+            if (type.type == Variant::COLOR) {
+                actions.push_back(
+                _script_node_builder<OScriptNodeDecompose>(
+                    category,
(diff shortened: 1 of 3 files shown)


Naros:
Backported to 2.3 and 2.2.

PR #1255, opened 2026-02-08
GH-1253 Resize node to content on autowire

### LINKED ISSUE #1253: Resize/Shrink node when autowired
### Describe the bug

In some cases, some default value widgets are particularly large, such as AABB, Projection, Basis, etc. But when dragging from a pin of that type and selecting a function or property that takes the instance as an input argument, the node's size is not shrunk post autowire, making the node look highly unusual, e.g.

Image

On autowire scenarios, we'd expect the node to initially be rendered as 

Image

Now of course, enabling auto-shrink nodes solves the issue, but there may be reasons why users don't want that enabled, but we should at least force auto-resize on autowire during the initial render. Afterward, resizing is driven by the user.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1253

### DIFF
--- a/src/editor/graph/graph_panel.cpp
+++ b/src/editor/graph/graph_panel.cpp
@@ -2081,29 +2081,11 @@ void OrchestratorEditorGraphPanel::_queue_autowire(OrchestratorEditorGraphNode* 
         return;
     }
 
-    if (choices.size() == 1) {
-        // When there is only one choice, there is no need for the autowire dialog.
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.3 and 2.2

PR #1258, opened 2026-02-08
GH-1257 Optimize script save & reload signals

### LINKED ISSUE #1257: Saving an Orchestration is slow when multiple scripts opened
### Describe the bug

When using the universal `Ctrl+S` to save the scene and all open files, the save progress box can take several seconds. This appears to be compounded by the number of Orchestration scripts and the number of open functions/graphs.

In addition, scripts with comment nodes will report an error with `move_child`.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

It appears the issue stems from several factors that converge to create it.

* First, `OScript` is configured to allow the editor to reload the content from a file. This should not be the case, as it is delegated to Orchestrator, much like GDScript delegates it to the script editor plugin. This can cause unexpected script reloads, often triggered by the editor's filesystem dock.
* The OrchestratorEditor save_resource_in_path function makes an additional call to the filesystem dock to rescan the script. This is unnecessary because Godot handles this automatically; it can be removed.
* The `OScript::reload` function is called multiple times through Godot's normal workflow and Orchestrator's. The function always reloads the Orchestration from disk, recreates it, and notifies the editor UI. This results in multiple UI redraws and reloads, generating a large number of signals that fire during save. This is the primary reason for the slow saves. On reload, we should only reload from disk if the contents have changed or the script is marked to force reload.

Naros:
Fixes #1257

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -2113,7 +2113,7 @@ void OrchestratorEditor::save_resource_in_path(const Ref<Resource>& p_resource, 
     }
 
     p_resource->set_path(p_path);
-    EI->get_resource_filesystem()->update_file(p_path);
+    // EI->get_resource_filesystem()->update_file(p_path);
 
     EditorNode->emit_signal("resource_saved", p_resource);
 }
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.3 and 2.2

PR #1264, opened 2026-02-12
GH-1263 Do not auto-init script language

### LINKED ISSUE #1263: Recent load changes causes language to load too early
### Describe the bug

We recently made a change in GH-1242 to deal with the script language not being initialized properly by Godot. The problem is we are now registering the language too early, before all singletons are registered, which breaks scripts that use specific singletons.

### Expected behavior

All singletons should be available.

### Actual behavior

Only a subset of singletons are accessible.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1263 

This forces the scripting language to rely on a forced editor restart when the language is enabled only.

### DIFF
--- a/src/api/extension_interface.cpp
+++ b/src/api/extension_interface.cpp
@@ -41,9 +41,6 @@ namespace orchestrator {
         if (p_level == MODULE_INITIALIZATION_LEVEL_EDITOR) {
             register_script_scene_types();
             register_editor_types();
-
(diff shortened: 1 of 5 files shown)


Naros:
Backported to 2.3 and 2.2.

PR #1265, opened 2026-02-12
GH-1262 Correctly reset override function state

### LINKED ISSUE #1262: Adding function call mistakenly added as a function override
### Describe the bug

When selecting a function call in the action dialog, it is registered as a function override rather than a call function node.

### Expected behavior

Should spawn a call function node.

### Actual behavior

Registers a script function.

### How to Reproduce?

1. Add a function override, like `_ready` or `_process`.
2. Right-click and select `AudioServer` singleton.
3. Drag from the singleton node output pin and select `get_bus_volume_db`.
4. Function is registered as a function override in the component panel, rather than a call function node.

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1262


Naros:
Backport to 2.3 and 2.2.

PR #1268, opened 2026-02-14
GH-1267 Add plugin-level autoload caching

### LINKED ISSUE #1267: Script parsing is slow when using many autoloads
### Describe the bug

During the script parsing, there is an autoload lookup that reparses the data from the Godot `ProjectSettings` API. While this isn't read from disk, the parsing adds extra latency.

### Expected behavior

Scripts with many references to autoloads should not be slower to parse.

### Actual behavior

_No response_

### How to Reproduce?

Create a script that refers to an autoload 24 or 48 times, particularly in a single function call. Then observe how long it takes to parse that function versus other functions that do not contain any autoload references.

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1267


Naros:
Backported to 2.3 and 2.2.

PR #1270, opened 2026-02-14
GH-1269 Enable compiler node convergence

### LINKED ISSUE #1269: Enable compiler feature: Node Convergence
### Description

The Orchestration compiler includes a feature called _Node Convergence_, which was initially disabled. This feature is a massive optimization to the bytecode generation pass, reducing the footprint for methods that heavily rely on branching and other multi-path control flow nodes.

### Implementation ideas

_No response_

Naros:
Fixes #1269


Naros:
Backported to 2.3 and 2.2.

PR #1276, opened 2026-02-14
GH-1275 Fix free object working with controls

### LINKED ISSUE #1275: Free Object does not properly handle Node-derived types
### Describe the bug

The goal behind the "Free Object" node is that users do not need to know the object's type, and that the scripting engine will resolve the best way to free the content, by either using `queue_free()` for Node-based types or `free` for non-Node types.

The current node only seems to work for the later.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1275

### DIFF
--- a/src/script/compiler/compiler.cpp
+++ b/src/script/compiler/compiler.cpp
@@ -19,6 +19,7 @@
 #include "common/dictionary_utils.h"
 #include "common/error_list.h"
 #include "core/godot/config/project_settings_cache.h"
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.3 and 2.2.

PR #1278, opened 2026-02-14
GH-1277 Omit math constants in debugger global list

### LINKED ISSUE #1277: Omit math constants in debugger global list
### Description

When GDScript breaks on a line, the math constants are not shown in the debugger global list. This is currently not the case when a breakpoint is observed in Orchestrator, and we should align this behavior to be consistent with GDScript.

### Implementation ideas

_No response_

Naros:
Fixes #1277


Naros:
Backported to 2.3 and 2.2

PR #1279, opened 2026-02-15
GH-1259 Disable support for built-in scripts

### LINKED ISSUE #1259: Disable Built-in Script Support
Given that there are current limitations for built-in script support, outlined in #1140, we should adjust the OScriptLanguage function `_supports_builtin_mode` to return false. 

This would prevent attempting to use Built-In Script functionality until its ready.

Naros:
Fixes #1259


Naros:
Backported to 2.3 and 2.2.

PR #1280, opened 2026-02-15
GH-1244 Update editor/inspector on variable/signal changes without an explicit save

### LINKED ISSUE #1244: Editor/Inspector does not update without an explicit save
### Describe the bug

One of the nice features of the script text editor is that periodically as you type, the script is reparsed and the various consumers of the script state reflect the latest changes. For example, if a variable is not exported but is annotated with `@export`, clicking back on the node with the script reflects the changes, even though the script was not explicitly saved.

In Orchestrator, while we do have periodic validation and updates that happen on specific graph actions, such as moving nodes, adding/removing nodes, or connecting/disconnecting nodes, not every action triggers the script validation pass, which is responsible for updating export variable state and more.

### Expected behavior

Any change should trigger a revalidation & editor interface update.

### Actual behavior

Not all actions trigger the expected behavior.

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1244

### DIFF
--- a/src/editor/script_components_container.cpp
+++ b/src/editor/script_components_container.cpp
@@ -51,6 +51,11 @@ Ref<Orchestration> OrchestratorScriptComponentsContainer::_get_orchestration() {
     return _orchestration;
 }
 
+void OrchestratorScriptComponentsContainer::_variables_changed() {
+    _set_edited(true);
+    _update_variables();
+}
+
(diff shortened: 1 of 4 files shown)


Naros:
Backported to 2.3 and 2.2.

PR #1294, opened 2026-02-21
Fix @OScript description for constant `One`

Naros:


Naros:
Backported to 2.3 and 2.2

PR #1300, opened 2026-02-22
GH-1299 Avoid text resource truncation

### LINKED ISSUE #1299: Text-based resource becomes truncated when editor calls `set_uid`
### Describe the bug

The text-based parser uses a 2 KB buffered input stream to optimize reads from disk. The problem is that when `rename_dependencies` or `set_uid` is called, the current API does not provide a way to know the read offset into the buffer, and so any data in the 2 KB buffer is lost when the remaining portion of the file is flushed to disk, leading to truncated resource files.

### Expected behavior

When `rename_dependencies` or `set_uid` is called, the resource should not be truncated.

### Actual behavior

_No response_

### How to Reproduce?

1. Create a text-based `.torch` script with some nodes and save it.
2. Tab out of the editor to the file system and make a copy of the script with a new name.
3. Tab back into the editor.
4. Editor crashes

After reviewing the content of the script copy in (2), we found that the script is truncated due to the buffered reads.

### Godot full version

_No response_

### Orchestrator version

2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1299

### DIFF
--- a/src/orchestration/serialization/binary/binary_serializer.cpp
+++ b/src/orchestration/serialization/binary/binary_serializer.cpp
@@ -563,6 +563,10 @@ bool OrchestrationBinarySerializer::recognize(const Ref<Resource>& p_resource) {
 }
 
 Error OrchestrationBinarySerializer::set_uid(const String& p_path, int64_t p_uid) {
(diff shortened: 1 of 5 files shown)


Naros:
It does not appear that 2.0 and 2.1 branches implement `set_uid`, so removing those from backports.

Naros:
Backported to 2.3 and 2.2

PR #1326, opened 2026-02-28
GH-1325 Fix `_init` method not being called

### LINKED ISSUE #1325: The _init method on scripts is no longer called.
### Describe the bug

When a script uses the `_init` method, it is no longer called.

### Expected behavior

The method should be called.

### Actual behavior

_No response_

### How to Reproduce?

Create a script with an `_init` method, adding any logic, such as a `prints` or even a `Print String` node. Attach the script to a node in a scene and run the scene.  The `_init` method is never fired.

### Godot full version

_No response_

### Orchestrator version

2.5/2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes #1325

### DIFF
--- a/src/script/language.cpp
+++ b/src/script/language.cpp
@@ -1395,7 +1395,7 @@ OScriptLanguage::OScriptLanguage() {
     _singleton = this;
     lock.instantiate();
 
-    strings._init = StringName("init");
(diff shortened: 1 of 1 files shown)


Naros:

PR #1331, opened 2026-02-28
GH-367 Make key bindings configurable

### LINKED ISSUE #367: Make key bindings configurable

### LINKED ISSUE #368: Consider adjusting the key bind for Place/Remove Knots

### LINKED ISSUE #735: German Keyboard Layout: `Ctrl+/` (Slash) does not toggle Component Panels
### Describe the bug

Using german Keyboard layout does not toggle the Componet Panels. With english Keyboard Layout it works. The slash (`/`) on the german Keyboard needs `Shift+7`, in combination with **Ctrl** that then will be `Ctrl+Shift+7`.

### Expected behavior

No matter which Keyboard layout is used the Keyboard shortcut should work.

### Actual behavior

With german Keyboard layout the shortcut `Ctrl+/` can not be executed.

### How to Reproduce?

1. Open Godot
2. Create new Orchestrator EventGraph
3. Use german Keyboard layout
4. Press `Ctrl+Shift+7` (`Ctrl+/`) >> Component Panels do not toggle

### Godot full version

4.3.rc3

### Orchestrator version

2.1.rc4

### Additional information

_No response_

Naros:
Fixes #367 
Fixes #735 
Fixes #368 

All settings moved from ProjectSettings to the new 
image

Settings dialog tab:
image

Shortcuts dialog tab:
image


Naros:
This likely cannot be backported to 2.3 and 2.2 easily, as the shortcut API on `EditorSettings` does not exist.

Naros:
Reworked the solution to be Godot version agnostic, so this should be backportable to 2.2 and 2.3, but requires some testing to verify the build works.

PR #1339, opened 2026-03-03
GH-709 Support connection straightening

### LINKED ISSUE #709: Add ability to straighten pins between nodes & reroute/knots
### Description

Users would like to have the ability to straighten connection wires between nodes and reroute nodes such that they make perfectly straight lines 

### Implementation ideas

_No response_

Naros:
Fixes #709 

Given this graph
image

The pin context menu offers these options:
image

When selecting straighten all connections, the result is:
image

In addition, of a connection has a knot, rather than the target node pin being aligned with the source pin, the first knot is aligned to the source pin only, which mirrors the behavior observed in UE.
image


Naros:

PR #1342, opened 2026-03-07
Style: Replace header guards with `#pragma once`

Naros:


Naros:

PR #1374, opened 2026-03-16
GH-1365 Avoid unnecessary component panel refreshes

### LINKED ISSUE #1365: Cannot create a variable more than once or until at a point in an empty `Orchestration`
### Describe the bug

Honestly, very simple. If trying to create a variable two or multiple times in a row in an empty `Orchestration`, it removes the field for the new variable right after showing it. And trying to press the + button to create a variable does not work.

### Expected behavior

It is supposed to do as you would create a variable before the bug happens, to show the new variable field then allow to name it right after creating. Then use it of course.

### Actual behavior

After trying multiple times to create a variable, it does exactly what I said in Describe the bug.

### How to Reproduce?

Reproducer :

Steps to reproduce (in an empty Orchestrator. Should be able to be reproduced in any Godot project but personally, I did in an empty one) :

1. Create an empty Orchestrator. Just click on `New Script`, select `OScript` for the language then press `Create`.
2. Try to create a new variable multiple times. The type or name does not matter, just press Enter after creating.
3. At a point, the bug should appear.

### Godot full version

v4.6.1.stable.arch_linux [14d19694e]

### Orchestrator version

v2.4.2.stable.official

### Additional information

- Using `Promote to Variable` does not work as a workaround. The field does not even show but does when restarting the Godot project.
- In Orchestrator v2.4.0, the bug started to exist and actually, trying the workaround above was indeed working and shows the field. But then trying to edit the variable just crashes Godot editor.
- It sometimes reports the error `ERROR: No item selected.`

Naros:
Fixes #1365

### DIFF
--- a/src/editor/graph/graph_panel.cpp
+++ b/src/editor/graph/graph_panel.cpp
@@ -2356,6 +2356,10 @@ void OrchestratorEditorGraphPanel::_drop_data_files_as_exported_variables(const 
         variable->set_exported(true);
         variable->set_info(property);
         variable->set_classification("class:" + res->get_class());
+
+        emit_signal("nodes_changed");
+
+        _set_edited(true);
     }
 }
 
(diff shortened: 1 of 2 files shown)


Naros:
Backported to 2.4 only, was unnecessary for 2.3 and 2.2.

Issue #1392, opened 2026-03-20
Virtual method overrides does not include `_get` / `_set`


Naros:
### Describe the bug

The `_get` and `_set` methods should be accessible for direct object manipulation of values.

### Orchestrator version

2.5/2.4/2.3/2.2

PR #1393, opened 2026-03-20
GH-1392 Allow overriding `_set` and `_get`

Naros:
Fixes GH-1392


Naros:

PR #1660, opened 2026-06-21
Reduce artifact build sizes

Naros:


Naros:
Appears that this reduces the size of the artifacts for Linux x86_64/arm64 and Android 32/64 only.

PR #1700, opened 2026-07-04
GH-1683 Fix reload on external changes to work like ScriptEditor

### LINKED ISSUE #1683: Reload from disk does not clear edited flag
### Describe the bug

When a script is modified outside the editor, pressing "Reload from disk" on the pop-up does not clear the edited flag.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Godot full version

_No response_

### Orchestrator version

2.5/2.4/2.3/2.2

### Additional information

_No response_

Naros:
Fixes GH-1683

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -66,6 +66,7 @@
 #include <godot_cpp/classes/scene_tree.hpp>
 #include <godot_cpp/classes/script_editor.hpp>
 #include <godot_cpp/classes/texture_rect.hpp>
+#include <godot_cpp/classes/thread.hpp>
 #include <godot_cpp/classes/v_separator.hpp>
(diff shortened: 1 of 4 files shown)


Naros:
The logic in `Script::reload_from_file()` differs between Godot 4.7 and 4.6. This solution may not work for 4.6 because rather than delegating reload to the `ScriptLanguage`, it gets delegated to the `Script` for non-tools.

Naros:
This definitely does not work as expected on 2.4/Godot 4.6.

Naros:
Due to the engine limitations, this will only be fixed on Godot 4.7+, which is Orchestrator 2.5+.

